PR #90, opened 2016-09-03
WIP: Adds CLI Tests

ccpricenytimes:


ccpricenytimes:
Closing this in favor of Jest tests.

We still should have some end to end tests. I'll create an issue to address it.

tizmagik:
@ccpricenytimes I don't think we need to lose the work you've done here. We should be able to refactor these tests in Jest pretty easily. Thoughts?

PR #115, opened 2016-09-09
Jest

tizmagik:
Uses [Jest](https://facebook.github.io/jest/) as a test runner, instead of AVA.

Jest supports in-memory transpiling of source and test files (via Babel) which removes the expensive step of transpiling to disk and thus significantly speeds up test runtime. It also comes with some very impressive features such as a changed-only watch mode, [snapshot testing](https://facebook.github.io/jest/blog/2016/07/27/jest-14.html), and out of the box code coverage reporting.

#### AVA Runtime
image

#### Jest runtime at "cold boot"
image

#### Jest runtime in "watch" mode
image

_There's an extra test for Jest because I added a dummy test to get the harness working._

---

Still todo:

- [x] ~~Potentially refactor how babel config integrates with userland. Right now, this code reuses `modifyWebpackConfig()` to grab the appropriate "development" babel config.~~ To be done in #134 
- [x] Remove AVA and clean up other artifacts
- [x] Add `--watch` support in the kyt CLI (Jest supports this out of the box)
- [x] Add `--coverage` support
- [x] Documentation


.eslintrc
```diff
@@ -21,7 +22,7 @@
       "no-warning-comments": [1, { "terms": ["todo", "fixme"], "location": "start" }],
       "react/sort-comp": 0,
       "react/jsx-filename-extension": [1, { "extensions": [".js", ".jsx"] }],
-      "import/no-unresolved": [2, { "ignore": ["ava"] }],
+      "import/no-unresolved": 2,
```
delambo: I think airbnb already has this configured, so we can remove it now.

docs/kytConfig.md
```diff
@@ -15,29 +15,46 @@ and export an object with the following options.
...
+There are no options passed in (like how `modifyWebpackConfig` has) since the only option would be a "test" environment, which would be redundant.
+
+```javascript
+modifyJestConfig: (baseConfig) => {
+  // modify baseConfig as needed
+  return baseConfig;
+}
+```
+
+> **NOTE:** Currently the base Jest config runs `modifyWebpackConfig()` for you with `options.environment = "development"` in order to grab the development settings for your app (babel plugins, aliases, etc). For most cases, this will be sufficient configuration to get your tests working out of the box without a need to define `modifyJestConfig` yourself. So it is recommended that you try running the tests through first and only add this callback if it's needed. If you try it out and this assumption doesn't seem to work for your use case, please file an issue.
```
tizmagik: I think it's important that we note to users that `modifyWebpackConfig()` still runs as there's no way of knowing that otherwise, but I can stop at that first example and provide a recipe for the Jest config.

ccpricenytimes: Oh yeah I just meant to add recipes

tizmagik: Oh okay, should I restore the comment? I added a note in the recipes as you suggested, let me know what you think!

ccpricenytimes: I think what's there now is good. thanks

utils/jest/preprocessor.js
```diff
@@ -0,0 +1,22 @@
...
+
+// Merge userland babel config with our babel config
+// TODO: This logic should probably move to config/babel.js,
+// but that would be a breaking change.
+const clientBabelConfig = clientConfig.module.loaders
+  .find(loader => loader.loader === 'babel-loader')
+  .query;
+
+// TODO: What's sort of confusing here is that userland babel config
+// (at least for Project-Vi) does not put things in babel.env's.
```
delambo: I know this is kind of a small thing, but can we leave references to internal NYT tools outside of kyt now that it's open? Outside users won't understand what we're comparing against. Possibly even make this an issue instead of a comment/aside?

tizmagik: Yea for sure! We actually already have an issue to address this #134 so I'll delete this comment. Good catch!

delambo:
Thanks for updating the diagram! 🙏 🙏 🙏

utils/jest/preprocessor.js
tizmagik: Addressed.

tizmagik:
@delambo Thanks! Reminder to also merge the corresponding PRs in the starter-kyts:
https://github.com/NYTimes/kyt-starter/pull/5
https://github.com/NYTimes/kyt-starter-universal/pull/3

PR #156, opened 2016-09-14
Unit tests

### LINKED ISSUE #133: Unit tests for kyt

tizmagik:
Fixes #133 

image

Relies on #115 

It's about as close as I could get it. This was a great exercise in learning Jest! 😃 

- `setup.js` needs to be refactored: #159 
- `test.js` can't be tested, I don't think, because inception. 
- And `commands.js` is difficult or impossible to test because of the way commanderjs works, but should be okay because of the e2e tests.

Still TODO:

- [x] Make eslint happy, specifically need a solution for `no-global-require` rule.


package.json
```diff
@@ -52,7 +55,7 @@
     "glob": "7.0.6",
     "inquirer": "1.1.2",
     "install": "0.8.1",
-    "jest": "15.1.1",
+    "jest": "^15.1.1",
```
ccpricenytimes: no ^ here

utils/kytConfig.js
```diff
@@ -4,12 +4,12 @@
 const path = require('path');
 const shell = require('shelljs');
 const mergeAll = require('ramda').mergeAll;
+const logger = require('../cli/logger');
```
ccpricenytimes: We can't do this. It's a chicken and egg problem. The logger references kytConfig. So in this file we just use the console.

cli/__tests__/.eslintrc
```diff
@@ -0,0 +1,5 @@
+{
```
ccpricenytimes: what is this?

cli/__tests__/logger.test.js
```diff
@@ -0,0 +1,60 @@
...
+
+  const simpleMethods = [
+    { method: 'log', expectedPrefix: '' },
+    { method: 'task', expectedPrefix: '👍  ' },
+    { method: 'start', expectedPrefix: '🔥  ' },
+    { method: 'end', expectedPrefix: '✅  ' },
+    { method: 'info', expectedPrefix: 'ℹ️  ' },
+  ];
+
+  const advancedMethods = [
```
ccpricenytimes: I think calling these simple and advanced isnt' really the distinction. It's about whether they get passed extra data yes? 

Maybe textCommands objectCommands or something like that

cli/actions/__tests__/build.test.js
```diff
@@ -0,0 +1,74 @@
+const shell = {
```
ccpricenytimes: should we move this somewhere more global? seems like a lot of the files would need this mock

cli/actions/__tests__/lint.test.js
```diff
@@ -0,0 +1,39 @@
...
+  const lint = require('../lint');
+  const eslint = require('eslint');
+
+  beforeEach(() => {
+    jest.resetModules();
+  });
+
+  it('logs and exits 0 when no errors', () => {
+    lint();
+    expect(logger.end).toBeCalledWith('Your JS looks great ✨');
```
ccpricenytimes: can we test partial strings with this? seems brittle to include emojis and everything

utils/__mocks__/ifPortIsFreeDo.js
```diff
@@ -0,0 +1,3 @@
+const mock = jest.fn((port, todo) => { todo(); });
```
ccpricenytimes: This may be silly but can we put __mocks__ inside their __tests__ folders? 
And then any that would be used in the top level could be in __mocks__ at the root?

utils/kytConfig.js
tizmagik: Take a look at logger, it no longer references kytConfig. This way logger stays 'pure' and the logic around logging or not is embedded in the app.

cli/__tests__/.eslintrc
tizmagik: To get around the no-global-require rule tripping since those would be in basically every test.

cli/__tests__/logger.test.js
tizmagik: @jaredmcdonald

utils/__mocks__/ifPortIsFreeDo.js
tizmagik: By Jest's convention a mock lives adjacent to the module it is mocking, with the exception of a `node_modules` mock which would live at the top-level `__mocks__` directory.

https://facebook.github.io/jest/docs/manual-mocks.html#content

cli/actions/__tests__/build.test.js
tizmagik: There is a global shell mock, but build needed a slightly different behavior, so it provides its own here.

ccpricenytimes: ok

package.json
tizmagik: That's a merge issue, it's not in `feature/jest` branch. I'll resolve, thanks!

cli/actions/__tests__/lint.test.js
tizmagik: We could but then it would look something like this:

```js
expect(logger.end.mock.calls[0][0]).toMatch(/Your JS looks great/)
```

Which would be potentially even more brittle depending on if `logger.end` was called anywhere else (`.mock.calls[x][x]` would be variable).

I think this is probably okay for now and if it ever fails we can relax this assertion or remove it entirely and assert something more useful.

ccpricenytimes: ok

utils/__mocks__/ifPortIsFreeDo.js
ccpricenytimes: ok

utils/kytConfig.js
ccpricenytimes: hmm when did that change?

tizmagik: in https://github.com/NYTimes/kyt/pull/127

delambo:
👍  x 801

PR #165, opened 2016-09-15
chore: fix for deprecated eslint rule

### LINKED ISSUE #164: react/require-extension is deprecated
```
The react/require-extension rule is deprecated. Please use the import/extensions rule from eslint-plugin-import instead.
```

cwonrails:
Closes #164 

Replaces the deprecated [`react/require-extension`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/require-extension.md) in `.eslintrc` with the recommended [`import/extensions`](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/extensions.md) rule from `eslint-plugin-import`.

Additionally, applies [this fix](https://github.com/airbnb/javascript/issues/978#issuecomment-244611697) to prevent the following (now incorrect) console output: `The react/require-extension rule is deprecated. Please use the import/extensions rule from eslint-plugin-import instead.`  

See [these lines](https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb-base/rules/imports.js#L109-L110) in `eslint-config-airbnb` for the relevant TODO.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -20,7 +20,8 @@
       "prefer-reflect": 2,
       "no-warning-comments": [1, { "terms": ["todo", "fixme"], "location": "start" }],
       "react/sort-comp": 0,
-      "react/jsx-filename-extension": [1, { "extensions": [".js", ".jsx"] }],
+      "react/require-extension": "off",
+      "import/extensions": [1, { "js": "never", "jsx": "never" }],
       "import/no-unresolved": [2, { "ignore": ["ava"] }],
       "import/no-extraneous-dependencies": [0],
       "no-param-reassign": 0
--- a/.eslintrc
+++ b/.eslintrc
@@ -21,7 +21,7 @@
       "no-warning-comments": [1, { "terms": ["todo", "fixme"], "location": "start" }],
       "react/sort-comp": 0,
       "react/require-extension": "off",
-      "import/extensions": [1, { "js": "never", "jsx": "never" }],
+      "import/extensions": [1, { "js": "never" }],
       "import/no-unresolved": [2, { "ignore": ["ava"] }],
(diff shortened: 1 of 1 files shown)


tizmagik:
Thanks @cwonrails ! We'll need to remember to follow up on this when https://github.com/benmosher/eslint-plugin-import/issues/390 lands.

.eslintrc
```diff
@@ -20,7 +20,8 @@
       "prefer-reflect": 2,
       "no-warning-comments": [1, { "terms": ["todo", "fixme"], "location": "start" }],
       "react/sort-comp": 0,
-      "react/jsx-filename-extension": [1, { "extensions": [".js", ".jsx"] }],
+      "react/require-extension": "off",
+      "import/extensions": [1, { "js": "never", "jsx": "never" }],
```
ccpricenytimes: You can remove jsx here since kyt doesn't support that extension.

cwonrails:
Definitely, thanks! Given that the version of `eslint-plugin-import` in `package.json` is now outdated, would a separate PR bumping the outdated dependencies be in order? 
screen shot 2016-09-15 at 4 46 54 pm

.eslintrc
cwonrails: Will do.

cwonrails: Fix pushed, let me know if I should squash or amend the original instead based on the Times' preference.

tizmagik: We squash & merge PRs, so it should be fine.

ccpricenytimes: I can squash when we merge from github so separate commits is fine

cwonrails: Great, thanks!

PR #167, opened 2016-09-20
fixes bug where dev server doesnt close after webpack compile error

### LINKED ISSUE #51: kyt dev occasionally fails to restart
I don't have specific steps for recreating this but occasionally when I'm developing (with `kyt dev`) the server fails to restart, saying the port is already in use:

```
👍  Client build successful
webpack built 7a148d7c5e66a4dae11c in 2544ms
👍  Server build successful
👍  Development server restarted
events.js:160
      throw er; // Unhandled 'error' event
      ^

Error: listen EADDRINUSE :::3000
```

killing the process seems to release the port again.

(glancing thru other issues - maybe related to #16?)

ccpricenytimes:
Fixes #51 

The dev server wasn't closing when exiting dev after a Webpack compile error. Now the server exits cleanly.

### DIFF
--- a/cli/actions/dev.js
+++ b/cli/actions/dev.js
@@ -18,6 +18,11 @@ const { buildPath, serverSrcPath } = require('../../utils/paths')();
 module.exports = (config) => {
   logger.start('Starting development build...');
 
+  // Kill the server on exit.
+  process.on('SIGINT', () => {
+    process.exit();
+  });
+
   let clientCompiler;
   let serverCompiler;
   const { clientConfig, serverConfig } = buildConfigs(config);
--- a/cli/actions/dev.js
+++ b/cli/actions/dev.js
@@ -19,9 +19,7 @@ module.exports = (config) => {
   logger.start('Starting development build...');
 
   // Kill the server on exit.
-  process.on('SIGINT', () => {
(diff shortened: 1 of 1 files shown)


delambo:
Can you ELI5?

ccpricenytimes:
So if someone started up dev and received a Webpack compile error they would see an error message. This prompts some users to ^C in their terminal to stop the dev process and make their changes. But this did not kill the underlying process so when restarting the dev server there was a port in use error (more details in @tizmagik comment in #51) What I've done here is created a way to capture that ^C when the user wants to exit and I shut down the process which cleanly exits the running servers. 

Does that make sense?

cli/actions/dev.js
```diff
@@ -18,6 +18,11 @@ const { buildPath, serverSrcPath } = require('../../utils/paths')();
 module.exports = (config) => {
   logger.start('Starting development build...');
 
+  // Kill the server on exit.
+  process.on('SIGINT', () => {
```
tizmagik: This can be

```javascript
process.on('SIGINT', process.exit);
```

ccpricenytimes: ok

PR #170, opened 2016-09-21
Linter refactoring and setup bug

### LINKED ISSUE #91: Extend kyt's local linters when setting up eslint and stylelint
The way kyt handles linter configurations isn't ideal. Currently, on `setup`, we drop the linter rc's into the user's root directory so the user can extend them and use them in editor plugins. Since the linters are no longer packaged, kyt has no way to update those files. 

To work around this, I propose we separately package the linter configurations inside of kyt and push them up to npm. That way, on `setup`, we can continue giving user's extensible linter rc's, but also gain the ability to update rules and dependencies in future releases.

Thoughts?

### LINKED ISSUE #166: Update .eslintrc to .eslintrc.json
.eslintrc is deprecated :( 

We should do several things. 

1) support all eslint config file types http://eslint.org/docs/user-guide/configuring#configuration-file-formats
2) on setup the file copied should be .eslintrc.json

delambo:
- adds kyt base linter configs in `configs/` which user linter configurations will `extend` - fixes #91 
- adds .json extensions to lintrc filenames - fixes #166 
- refactors user setup files into `config/user`
- fixes bug in `setup.js` where `tempPackageJSON` was undefined in `addPackageJsonScripts` during basic setup

One unfortunate thing about this change is that you can't locally lint with a global/linked kyt package: `kyt lint`.


cli/actions/lint.js
```diff
@@ -2,16 +2,25 @@
 // Command to lint src code
 
 const CLIEngine = require('eslint').CLIEngine;
+const path = require('path');
 const logger = require('./../logger');
 const glob = require('glob');
 const { userRootPath } = require('../../utils/paths')();
 
 module.exports = () => {
+  const eslintrc = glob.sync(`${userRootPath}/.*eslintrc*`);
```
tizmagik: Should we also look for `eslintConfig` in package.json?

cli/actions/lintStyle.js
```diff
@@ -10,15 +11,18 @@ module.exports = () => {
     process.exit(1);
   };
 
-  // Check to see if stylelint file exists
-  const stylelintrc = glob.sync(`${userRootPath}/.stylelintrc`);
-  if (!stylelintrc.length) {
-    handleError('You do not have a .stylelintrc file in the root of your project');
-  }
+  // Check to see if stylelint file exists and use base config as a backup.
+  const stylelintrc = glob.sync(`${userRootPath}/.stylelintrc.json`);
```
tizmagik: Similarly, `stylelint` property in user's `package.json`?

cli/actions/setup.js
```diff
@@ -152,60 +152,45 @@ module.exports = (config, program) => {
...
-    } else {
-      // Copy our .stylelintrc into the user's directory
-      const stylelintPath = path.join(__dirname, `../../config/${stylelintFileName}`);
-      if (shell.cp(stylelintPath, userStylelintPath).code === 0) {
-        logger.task('Copied default Stylelint config');
-      }
+    // Copy our .stylelintrc into the user's directory
+    const stylelintPath = path.join(__dirname, `../../config/user/${stylelintFileName}`);
+    if (shell.cp(stylelintPath, userStylelintPath).code === 0) {
+      logger.task(`Created ${stylelintFileName} file`);
```
tizmagik: Should we also log the error case?

cli/actions/setup.js
```diff
@@ -152,60 +152,45 @@ module.exports = (config, program) => {
...
-    } else {
-      // Copy our local eslint
-      const esLintPath = path.join(__dirname, '../../.eslintrc');
-      if (shell.cp(esLintPath, linkedPath).code === 0) {
-        logger.task('Copied kyt default ESLint config');
-      }
+    // Copy our user eslintrc into the user's root.
+    const esLintPath = path.join(__dirname, '../../config/user/.eslintrc.base.json');
+    if (shell.cp(esLintPath, linkedPath).code === 0) {
+      logger.task(`Created ${eslintFileName} file`);
```
tizmagik: What happens in the error case?

cli/actions/setup.js
```diff
@@ -234,15 +220,15 @@ module.exports = (config, program) => {
...
       shell.cp(newConfig, userKytConfigPath);
-      logger.task('Created new kyt.config.js');
+      logger.task(`Created ${configFileName} file`);
     };
 
     if (shell.test('-f', userKytConfigPath)) {
       // Since the user already has a kyt.config,
       // we need to back it up before copying.
-      const mvTo = path.join(userRootPath, `kyt.config-${Date.now()}.js.bak`);
+      const mvTo = path.join(userRootPath, `${configFileName}-${Date.now()}.bak`);
```
tizmagik: It might make sense to add a const at the very beginning for `Date.now()` so that the times are the exact same for all the backups.

config/user/.eslintrc.base.json
```diff
@@ -0,0 +1,14 @@
+{
+  "extends": [
+    "./node_modules/kyt/config/.eslintrc.base.json"
```
tizmagik: Why `.eslintrc.base.json` and not just `.eslintrc.json` ?

config/user/.stylelintrc.json
```diff
@@ -0,0 +1,5 @@
+{
+  "extends": "./node_modules/kyt/config/.stylelintrc.json",
```
tizmagik: For parity with eslint config I would expect to see `.stylelintrc.base.json`

docs/conventions.md
```diff
@@ -49,9 +49,9 @@ The `kyt test` command finds all files with `*.test.js` in the `/src` directory.
 
 The `kyt lint` command finds all files with `.js` extension in `/src`
 
-Specify lint rules with a `.eslintrc` file in the root of your project.
+Specify lint rules with a `.eslintrc.json` file in the root of your project.
```
tizmagik: What happens in the case where the user already has an `.eslintrc.json` (and also for `.styleintrc.json`)? My understanding is their file would get backed up and a new kyt-version would be created instead? It's not clear reading this that setup takes care of that for the user but changes can be made in that file after setup. Maybe that's out of scope for this specific doc, but it might be worth mentioning?

cli/actions/lint.js
delambo: I think we've covered enough use cases. We support all of the valid linter rc filenames and we document the filename convention. Not to mention most projects will use `setup` so they should be set up with the files.

cli/actions/setup.js
delambo: Added an error log.

cli/actions/setup.js
delambo: Added an error log.

cli/actions/setup.js
delambo: Good call. Added.

config/user/.eslintrc.base.json
delambo: eslint respects local eslintrc files. I tried `root:true` in the root file but that didn't seem to help.

docs/conventions.md
delambo: I updated the docs. In this section I mention and link to `setup` and then in the setup command documentation, I mentioned that we backup.

PR #171, opened 2016-09-21
Remove allChunks: true

### LINKED ISSUE #169: Remove `allChunks: true`
In https://github.com/NYTimes/kyt/blob/master/config/webpack.prod.client.js#L55

According to the docs, this flag forces all the CSS to be embedded in the entry CSS instead of in the individual chunks, see here: https://webpack.github.io/docs/stylesheets.html#all-styles-in-separate-css-output-file

tizmagik:
Fixes #169

### DIFF
--- a/config/webpack.prod.client.js
+++ b/config/webpack.prod.client.js
@@ -52,7 +52,7 @@ module.exports = options => ({
   },
 
   plugins: [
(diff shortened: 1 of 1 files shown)


tizmagik:
Closing this for now, will re-open after some more investigation. Bundle chunking/code-splitting isn't behaving as I expected it to and we might need to move some more things around.

PR #172, opened 2016-09-21
Use KYT for globals instead of process.env

coolov:
Currently `kyt` is replacing `process.env` with it's own global constants. This is breaking node apps that rely on environment variables for configuration. 

This PR is introducing `KYT` as a namespace for "compile time constants" that are defined by `kyt` (SERVER_PORT, CLIENT_PORT, PUBLIC_PATH, PUBLIC_DIR, ASSETS_MANIFEST).

It is a breaking update. Any app that is currently using for example `process.env.ASSETS_MANIFEST`, need to use `KYT.ASSETS_MANIFEST` instead.

### DIFF
--- a/docs/conventions.md
+++ b/docs/conventions.md
@@ -28,14 +28,13 @@ All `.js` files in `/src` are transpiled with Babel.
 
 ## Environment Variables
 
-kyt sets several environment variables with useful information about the app environment.
+kyt sets several global variables with useful information about the app environment.
 
-* `NODE_ENV` Node environment currently running (eg. production)
-* `SERVER_PORT` Port your node server should listen on.
-* `CLIENT_PORT` Port the client assets server is listening on.
-* `PUBLIC_PATH` Full path for static assets server
-* `PUBLIC_DIR` Relative path to the public directory
-* `ASSETS_MANIFEST` Object with build assets paths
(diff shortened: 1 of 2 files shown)


tizmagik:
LGTM, but do we need to mention this in the docs somewhere?

coolov:
yea, it needs to be in the docs. good catch @tizmagik

docs/conventions.md
```diff
@@ -28,14 +28,13 @@ All `.js` files in `/src` are transpiled with Babel.
...
-* `ASSETS_MANIFEST` Object with build assets paths
+kyt sets several global variables with useful information about the app environment.
+
+* `KYT.SERVER_PORT` Port your node server should listen on.
+* `KYT.CLIENT_PORT` Port the client assets server is listening on.
+* `KYT.PUBLIC_PATH` Full path for static assets server
+* `KYT.PUBLIC_DIR` Relative path to the public directory
+* `KYT.ASSETS_MANIFEST` Object with build assets paths
 
 For examples of how to use these environment variables, checkout out the simple React [starter-kyt](https://github.com/nytimes/kyt-starter)
```
coolov: @delambo --- it looks like the `starter-kyt` needs an update as well

coolov: i'll file a ticket

PR #174, opened 2016-09-22
serverless option

### LINKED ISSUE #151: Make server optional
An issue was raised https://github.com/NYTimes/kyt/issues/99 for making server HMR optional and I commented that we should make the server side optional altogether. I think that these are separate issues so I'm opening a new issue.

I'd love to see a POC that makes the server optional so that kyt can can manage a client side app that works against a separate server/api. This could be configured in the kyt.config.js.

jaredmcdonald:
(fixes #151)

- adds `hasServer` option, default: `true`
- branches logic in `build`, `start` and `dev` based on `hasServer`
- adds unit tests for new branches (and refactors them so that `beforeEach` hooks work)


delambo:
What do you think about using `hasServer` instead of `noServer`? Also, adding it to the `baseConfig` in utils/kytConfig.js as a default set to `true`.

cli/actions/dev.js
```diff
@@ -63,26 +63,32 @@ module.exports = (config) => {
...
+    if (!noServer) {
+      compileServer();
+    }
   });
 
   // Compile Server Webpack Config
-  serverCompiler = webpackCompiler(serverConfig, once(() => {
-    ifPortIsFreeDo(serverURL.port, startServer);
-  }));
+  if (!noServer) {
```
delambo: Maybe we should add the watcher code above into this block so we can group the features under one `if` statement.

delambo:
How have you been testing this? I'm thinking about making a quick starter-kyt that either runs on a static html file or a separate server.

delambo:
It's still a WIP, but I created a static starter-kyt to test this pull against: https://github.com/delambo/kyt-starter-static

The cool thing about the static starter-kyt is that it is forked from the universal starter-kyt so it will be pretty easy to maintain.

I think the no-server option will be useful in many use cases outside of the static starter kyt. That said, we might want to consider a convention for no-server, static file handling. For example, we could internalize the [html-webpack-plugin configuration](https://github.com/delambo/kyt-starter-static/blob/master/kyt.config.js#L10) in kyt and make it a convention that if the no-server option is on and there is a `src/index.ejs` file then we'll configure the html-webpack-plugin. Thoughts?

We should create another toy project that runs a Ruby/Go/whatever backend/api to test more use cases for this feature.

jaredmcdonald:
@delambo - implemented your feedback, want to take another look? this should be nearly mergeable now. (I was testing e2e with a local test project and with unit tests.)

re: other use cases, would be interesting to spin up a Rails project or something and try to use `hasServer: false` there... thinking we could put in further optimizations as future issues/PRs?

docs/commands.md
```diff
@@ -75,19 +75,22 @@ node_modules/.bin/kyt setup -r git@github.com:delambo/kyt-starter-universal-angu
 ## dev
 
 The `dev` command takes the entry index.js in `src/client/` and `src/server/`, compiles them, and starts client and backend servers. The dev environment includes hot reloading to allow for fast development.
+
+If `hasServer` is set to `false` in [kyt.config.js](/docs/kytConfig.md)), `src/server/` is ignored and no backend server is started.
```
ccpricenytimes: extra )

PR #184, opened 2016-09-25
WIP: Exits process on build error for CLI

### LINKED ISSUE #173: kyt build should pass along the exit code on build failure
On a build failure it seems the process still exits `0` which wouldn't break CI as expected

ccpricenytimes:
fixes #173 

Wanted someone else's opinion on this before I finalize it and fix the tests. Basically we want to different types of output from the Webpack compiler depending on whether its the dev or build task. The build task should exit so the CLI can register the error. The dev command should keep running so a user can make changes without having to restart the command manually.


utils/webpackCompiler.js
```diff
@@ -19,6 +19,11 @@ module.exports = (webpackConfig, cb) => {
     if (stats.hasErrors()) {
       logger.error(`${type} build failed\n`, stats.toString());
       logger.info('See webpack error above');
+
+      // Quit process if we're building
+      if (action === "build") {
+        process.exit(1);
+      }
```
tizmagik: The callback `cb` is already being called with the `stats` object, so instead of having to pass in an extra parameter,`action` here, I think it might be cleaner to just have the upstream code and the process.exit (e.g. check for `stats.hasErrors()` in the `dev` action). This way the process management stays in the actions and the underlying utility functions stay pure. What do you think?

ccpricenytimes:
@tizmagik this should be ready for a second look. let me know what you think!

PR #190, opened 2016-09-28
Adds newlines to log statements

delambo:
This is pretty small. I added some newlines to the `start`, `end`, and `error` logs. The `dev` command looks like the following after a hot reload:

```

ℹ️  Using kyt config at /Users/202586/Sites/test.nytimes.com/kyt-starter-universal/kyt.config.js

🔥  Starting development build...

👍  Cleaned ./build
👍  Client webpack configuration compiled
👍  Server webpack configuration compiled
👍  Client build successful
👍  Setup React Hot Loader
👍  Client assets serving from http://localhost:3001/assets/
webpack built 7741101ff391e9a5c1c4 in 4316ms
👍  Server build successful
👍  Server running at: http://localhost:3000/

✅  Development started

webpack building...
👍  Client build successful
webpack built a38681c47339c9e21496 in 1188ms
👍  Server build successful

✅  Development server restarted

```

### DIFF
--- a/cli/__tests__/logger.test.js
+++ b/cli/__tests__/logger.test.js
@@ -26,7 +26,8 @@ describe('logger', () => {
     it(`logger.${method}: simple usage`, () => {
       logger[method]('here is some text');
       if (newlineMethods.indexOf(method) > -1) {
-        expect(global.console.log.mock.calls).toEqual([[`${expectedPrefix}here is some text`],[]]);
+        expect(global.console.log.mock.calls)
+          .toEqual([[`${expectedPrefix}here is some text`], []]);
       } else {
         expect(global.console.log.mock.calls).toEqual([[`${expectedPrefix}here is some text`]]);
       }
@@ -39,7 +40,8 @@ describe('logger', () => {
       logger[method]('here is some text', { description: 'and a second argument' });
       expect(global.console.dir.mock.calls.length).toBe(0);
       if (newlineMethods.indexOf(method) > -1) {
-        expect(global.console.log.mock.calls).toEqual([[`${expectedPrefix}here is some text`],[]]);
+        expect(global.console.log.mock.calls)
+          .toEqual([[`${expectedPrefix}here is some text`], []]);
       } else {
         expect(global.console.log.mock.calls).toEqual([[`${expectedPrefix}here is some text`]]);
       }
@@ -52,7 +54,7 @@ describe('logger', () => {
       expect(global.console.dir.mock.calls.length).toBe(0);
       if (newlineMethods.indexOf(method) > -1) {
         expect(global.console.log.mock.calls).toEqual(
-          [[`${expectedPrefix}here is some text\nand a second string argument`],[]]
+          [[`${expectedPrefix}here is some text\nand a second string argument`], []]
         );
       } else {
         expect(global.console.log.mock.calls).toEqual(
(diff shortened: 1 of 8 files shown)


tizmagik:
What if instead of hard-coding the new-line char to only `start` `end` and `error` you handled it generically for all console logs to consume a log that starts with a new-line char and then pass along the rest of the call as normal? E.g.

```js
const write = (status, text, verbose) => {
    if (text.startsWith('\n')) {
        logger.log('');
        write(status, text.slice(1), verbose);
        return;
    }

    // ...
}
```

And that way, when some new-line "padding" is desired, you can just log like so:

```js
logger.end('\nDevelopment server restarted');
```

delambo:
My idea was to group tasks and to always separate the important boundary log statements like start and end. In general all logs would follow the following pattern:

```
Start

task 1
task 2
task 3

End
```

For long-running commands like `dev`, the restart tasks would be grouped and followed with an end log:

```
Start

task 1
task 2

End

restarting...
task 1
task 2

Error

restarting...
task 1
task 2

End

...
```

tizmagik:
I see, cool! I guess that makes sense for `start` and `end`, but why for `error`? `start` and `end` are not part of the canonical `console` API so it makes sense to add it as custom to the `logger` API, but I'm not sure that it makes sense to do that to `error`? What do you think?

cli/__tests__/logger.test.js
```diff
@@ -9,21 +9,28 @@ describe('logger', () => {
...
 
+  const newlineMethods = ['start', 'end', 'error'];
+
   [...oneArgMethods, ...twoArgMethods].forEach(({ method, expectedPrefix }) => {
     it(`logger.${method}: simple usage`, () => {
       logger[method]('here is some text');
-      expect(global.console.log.mock.calls).toEqual([[`${expectedPrefix}here is some text`]]);
+      if (newlineMethods.indexOf(method) > -1) {
+        expect(global.console.log.mock.calls)
+          .toEqual([[`${expectedPrefix}here is some text`], []]);
```
tizmagik: Could this be written as this?:

```js
expect(global.console.log).toBeCalledWith(`${expectedPrefix}here is some text`);
```

cli/__tests__/logger.test.js
```diff
@@ -9,21 +9,28 @@ describe('logger', () => {
...
+      } else {
+        expect(global.console.log.mock.calls).toEqual([[`${expectedPrefix}here is some text`]]);
```
tizmagik: I guess the same `toBeCalledWith()` could be used here as well and for the rest of this test, right?

jaredmcdonald: @tizmagik the difference here is that this also verifies it was called exactly once, with only that one argument... could add some more assertions to get that too, maybe this is too opaque

tizmagik: Yea, I think it's a bit too opaque, I think something like this is more descriptive:

```js
expect(global.console.log).toBeCalledWith(`${expectedPrefix}here is some text`);
expect(global.console.log.mock.calls).toEqual(1);
```

Or, when we update to `Jest v16`:

```js
expect(global.console.log).toBeCalledWith(`${expectedPrefix}here is some text`);
expect(global.console.log).toHaveBeenCalledTimes(1);
```

delambo: Fixed the tests up.

tizmagik:
Looks good! I still think we shouldn't add newlines for `logger.error()` because that would be unexpected, but I'll leave that up to you.

PR #220, opened 2016-10-05
add cheap source map to dev client Fixes #219

### LINKED ISSUE #219: Re-enable source maps in dev
Not sure when this started happening but currently the source-maps are not working at least in dev mode. For a production build, source maps are being emitted.

We should also probably use `devtool: 'cheap-module-eval-source-map'` or similar in dev.

leshow:


delambo:
I threw a quick test together which throws an error in one of our components (Accordion). The `cheap-module-eval-source-map` doesn't appear to show the component name at the top of the stack trace and when I follow the link to the source it takes me to the wrong file:

cheap-module-eval-source-map

Strangely, when I use `cheap-eval-source-map`, I get the component name in the error stack, but it takes me to a webpack compiled module which isn't very helpful:

cheap-eval-source-map

According to the webpack documentation, `cheap-module-eval-source-map` looks like what we want but it's not working well for me. What's going on? Maybe it has something to do with that component - I'll try another test later.

tizmagik:
Yea, I was seeing the same thing locally when I tried this same fix before opening the issue and went pretty deep down the rabbit hole. I'm not sure why the source-maps aren't working as expected.

leshow:
https://github.com/webpack/webpack/issues/2145

this is why I assume. any source-map that doesn't use the 'module' keyword appears to work as expected.

leshow:
Actually I stand corrected, it works with cheap-module-source-map.

I was able to set a breakpoint and use observe the source.

tizmagik:
Thanks @leshow , nice job digging up the upstream issue -- it seems it's a bug in Chrome after all. However `cheap-module-source-map` still doesn't seem to work for me, clicking log statement links in the debugger brings me to an `undefined ...` file in webpack.

@delambo let's track https://github.com/webpack/webpack/issues/2846 (which in turn tracks the upstream Chrome issue) and restore `cheap-module-eval-source-map` when it lands.

For now, I found that `inline-source-map` seems to work, but it's much slower, but maybe we can use that as a workaround for now, or try to experiment with other `devtool` values?

leshow:
inline-source-map works for me also. I've changed it to that and pushed. I think it's probably better to have slow source maps than no source maps at all for dev, they are essential for development.

delambo:
Thanks @leshow!

PR #222, opened 2016-10-06
Adds the public directory path as a dev server static directory

### LINKED ISSUE #223: Public assets not being served in dev
Files in /src/public aren't being served in Dev. 

We'd like them to serve from the client assets server.

Let's also export KYT.CLIENT_URL and KYT.SERVER_URL

delambo:
In `dev` the `src/public` directory assets aren't accessible. This configures the dev server to serve the public directory as a static directory. If you know of a better way to do this via webpack config, I'm all ears.

Fixes #223

### DIFF
--- a/__mocks__/express.js
+++ b/__mocks__/express.js
@@ -1,11 +1,14 @@
 const use = jest.fn();
 const listen = jest.fn();
+const statik = jest.fn();
+
 const expressMock = jest.fn(() => ({
   use,
   listen,
+  static: statik,
 }));
-
 
 module.exports = expressMock;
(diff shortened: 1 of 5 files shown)


tizmagik:
In the webpack config setting [output.publicPath](https://github.com/webpack/docs/wiki/configuration#outputpublicpath) doesn't do it?

delambo:
`publicPath` doesn't seem to work here. I think it defines the path for assets that are digested by webpack loaders.

delambo:
It turns out that there is a problem in our project, not kyt.

PR #230, opened 2016-10-08
Passes flags to commands

### LINKED ISSUE #113: Enable Debugging in kyt (Node Debugger in Chrome Dev tools)
I am looking for a way to run --debug=true with docker for project-vi.   

Here is how to run the debugger that I would like to run: 
https://medium.com/@paul_irish/debugging-node-js-nightlies-with-chrome-devtools-7c4a1b95ae27#.elnqyj305

### LINKED ISSUE #150: Pass through command flags that kyt doesn't use
We want to support other flags that kyt isn't using so developers can pass down command line options to tools we use. 

Example of use case here https://github.com/NYTimes/kyt/commit/d2e2ad878b3ee6ed32552ddf41251f266ab3cbf4#commitcomment-18973379

ccpricenytimes:
Fixes #150 
Fixes #113

### DIFF
--- a/cli/actions/__tests__/dev.test.js
+++ b/cli/actions/__tests__/dev.test.js
@@ -40,7 +40,9 @@ describe('dev', () => {
     clientURL: mockURL,
     serverURL: mockURL,
     reactHotLoader: false,
-  });
+  },
+  {},
+  []);
 
   it('cleans the build directory', () => {
     expect(logger.task).toBeCalledWith('Cleaned ./build');
@@ -83,9 +85,9 @@ describe('dev', () => {
   });
 
   it('server compilation done', () => {
+    const flagsStub = [];
     serverCompilerDone(stats);
-
-    expect(nodemon).toBeCalledWith({ script: 'fakePath', watch: ['fakePath'] });
+    expect(nodemon).toBeCalledWith({ script: 'fakePath', watch: ['fakePath'], nodeArgs: flagsStub });
 
     // nodemon.once
     expect(nodemon.once.mock.calls[0][0]).toBe('start');
--- a/cli/actions/__tests__/dev.test.js
+++ b/cli/actions/__tests__/dev.test.js
@@ -41,7 +41,6 @@ describe('dev', () => {
     serverURL: mockURL,
     reactHotLoader: false,
   },
-  {},
   []);
 
   it('cleans the build directory', () => {
--- a/cli/actions/dev.js
+++ b/cli/actions/dev.js
@@ -51,7 +51,7 @@ module.exports = (config, program, flags) => {
       serverCompiler.options.output.path, `${Object.keys(serverCompiler.options.entry)[0]}.js`
     );
 
-    nodemon({ script: serverPath, watch: [serverPath], nodeArgs:[...flags] })
+    nodemon({ script: serverPath, watch: [serverPath], nodeArgs: [...flags] })
       .once('start', () => {
         logger.task(`Server running at: ${serverURL.href}`);
         logger.end('Development started');
--- a/cli/actions/dev.js
+++ b/cli/actions/dev.js
@@ -15,7 +15,7 @@ const buildConfigs = require('../../utils/buildConfigs');
 const webpackCompiler = require('../../utils/webpackCompiler');
 const { buildPath, serverSrcPath } = require('../../utils/paths')();
 
-module.exports = (config, program, flags) => {
+module.exports = (config, flags) => {
   logger.start('Starting development build...');
 
   // Kill the server on exit.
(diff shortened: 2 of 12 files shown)


cli/actions/dev.js
```diff
@@ -51,7 +51,7 @@ module.exports = (config) => {
       serverCompiler.options.output.path, `${Object.keys(serverCompiler.options.entry)[0]}.js`
     );
 
-    nodemon({ script: serverPath, watch: [serverPath] })
+    nodemon({ script: serverPath, watch: [serverPath], nodeArgs: [...flags] })
```
tizmagik: Can just be `nodeArgs: flags`

cli/actions/__tests__/dev.test.js
```diff
@@ -40,7 +40,9 @@ describe('dev', () => {
     clientURL: mockURL,
     serverURL: mockURL,
     reactHotLoader: false,
-  });
+  },
+  {},
+  []);
```
tizmagik: Can we default these upstream so they don't need to be set if they're not needed?

cli/commands.js
```diff
@@ -22,8 +22,9 @@ if (!shell.test('-f', userPackageJSONPath)) {
 }
 
 const loadConfigAndDo = (action, optionalConfig) => {
+  const flags = program.args.filter(item => typeof item === 'string');
   const config = kytConfigFn(optionalConfig);
-  action(config, program);
+  action(config, program, flags);
```
tizmagik: It looks like `program` isn't actually used be any of the actions, and instead just need `flags`. Can you double check that, and if true, maybe we can simplify this signature to just this:

```js
action(config, flags);
```

ccpricenytimes:
@tizmagik Thanks for the feedback. Made the updates. setup is using `program` but I changed the order to make it a little cleaner

tizmagik:
Spent some time with this PR, still some issues I see:

This doesn't fully fix #150 e.g. we still can't do this:

```
kyt lint -- --fix
```

In that same vein, there is no distinction between `program.args` and `flags`. Ideally, we should be able to split the CLI flags on `--`, so that the following:

```
kyt cmd --arg1 --arg2 -- --flag1 --flag2
```

Produces:

```
args = ['arg1', 'arg2']; // we can probably drop "program" since that's an implementation detail
flags = ['flag1', 'flag2'];
```

---

If you want, we can say this fixes #113 but leave #150 open for a follow-up PR.

delambo:
I actually need this functionality right now for a starter kyt I'm building. I haven't fully formed a solution, but I think we could drop the eslint Node API and use the eslint cli. For example:

```
shell.exec(`${eslint} -c ${configFile} --ext .js ${USER_FLAGS} ${srcPath}`);
```

ccpricenytimes:
Hey @tizmagik, 

Ok more updates! Lint now takes any flags. I also separated flags and args. The way args works through commander is that they get parsed and placed into the command object, so I'm passing along that object to be used by setup, and creating the array of flags from anything that's a string. 

Let me know if you see anything else. Thanks!

@delambo test this out with your starter-kyt and let me know if it works

cli/actions/lint.js
```diff
@@ -1,42 +1,29 @@
...
   const lint = () => {
-    const files = ['src/'];
-    const cli = new CLIEngine(eslintCLI);
-    const report = cli.executeOnFiles(files);
-    const formatter = cli.getFormatter();
-    logger.log(`${formatter(report.results)}\n`);
-
-    if (report.errorCount === 0) {
-      logger.end(`Your JS looks ${report.warningCount === 0 ? 'great ✨' :
+    const cmd = `eslint src/ -c ${configFile} --color --env browser ${flags.join(' ')}`;
```
tizmagik: Could this just be `eslint src/ ${configFile}`?

docs/commands.md
```diff
@@ -77,6 +77,13 @@ node_modules/.bin/kyt setup -r git@github.com:delambo/kyt-starter-universal-angu
 The `dev` command takes the entry index.js in `src/client/` and `src/server/`, compiles them, and starts client and backend servers. The dev environment includes hot reloading to allow for fast development.
 Optionally, you can configure urls for the development servers in the [kyt config](/docs/kytConfig.md).
 
+You can pass flags to the node server through `kyt dev`.
+For example:
+```
+kyt dev -- --inspect
+```
+will run the [node debugging for Chrome DevTools](https://medium.com/@paul_irish/debugging-node-js-nightlies-with-chrome-devtools-7c4a1b95ae27#.mpuwgy17v)
+
```
tizmagik: Maybe we should also mention for the `test` and `lint` commands? Or, mention that flags can now be passed generically in almost all cases (except for `lint-style`) so it doesn't read as necessarily specific to these two commands?

cli/actions/lint.js
ccpricenytimes: we need to specify color and browser since we're running it with shelljs now

docs/commands.md
ccpricenytimes: I added separate notes. good catch!

cli/actions/lint.js
tizmagik: I think it might be better to put `env: { browser: true}` in the base eslint config instead, this way it's overridable at the command-line by the user.

I believe color is enabled by default anyway?

PR #235, opened 2016-10-08
adds modifyBabelConfig

### LINKED ISSUE #134: Refactor how Babel configuration is loaded in
This came up as part of the work on #115 . Right now, the way the Babel config is loaded for Jest is by running the `modifyWebpackConfig()` callback and grabbing it from there, but this overloads that method a bit. Now that there are 2 consumers for the Babel configuration (Webpack and Jest), it would be cleaner to pull that out.

Here's how it could work:
- Userland can optionally define a `.babelrc` file if they have any custom plugins or presets that they would like to use. (We should also check `package.json` for the configuration.)
- We'll respect the `BABEL_ENV` setting to pull in the appropriate dev/prod/test babel configuration.
- Merge it in with kyt's base babel configuration.

For most use cases, a separate babelrc file in userland won't be needed, but in the cases where it is, it will be easier to configure than having to do so programmatically in the `modifyWebpackConfig()` callback which requires sniffing for the `babel-loader`.

jaredmcdonald:
fixes #134 

adds a `modifyBabelConfig` callback in `kyt.config.js`

### DIFF
--- a/config/babel.js
+++ b/config/babel.js
@@ -1,5 +1,7 @@
 const path = require('path');
 const fs = require('fs');
+const config = require('../utils/kytConfig')();
+const logger = require('../cli/logger');
 
 // Uses require.resolve to add the full paths to all of the plugins
 // and presets, making sure that we handle the new array syntax.
@@ -26,5 +28,12 @@ module.exports = (options) => {
   }
   Object.keys(babelrc.env || {}).forEach(env => resolvePluginsPresets(babelrc.env[env]));
 
-  return babelrc;
+  // modify via userland kytConfig's `modifyBabelConfig`
+  try {
+    return config.modifyBabelConfig(babelrc, options);
+  } catch (error) {
+    logger.error('Error in your kyt.config.js modifyBabelConfig():', error);
+    // return to satisfy eslint `consistent-return` rule
+    return process.exit(1);
+  }
 };
--- a/config/babel.js
+++ b/config/babel.js
@@ -30,13 +30,13 @@ module.exports = (options) => {
   }
   Object.keys(babelrc.env || {}).forEach(env => resolvePluginsPresets(babelrc.env[env]));
 
-  // modify via userland kytConfig's `modifyBabelConfig`
   try {
-    const mergedBabelrc = config.modifyBabelConfig(babelrc, options);
+    // modify via userland kytConfig's `modifyBabelConfig`
+    const modifiedBabelrc = config.modifyBabelConfig(babelrc, options);
     if (config.debug) {
-      logger.debug('Merged babel configuration:', mergedBabelrc);
+      logger.debug('Modified babel configuration:', modifiedBabelrc);
     }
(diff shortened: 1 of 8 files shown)


config/babel.js
```diff
@@ -26,5 +30,16 @@ module.exports = (options) => {
...
   Object.keys(babelrc.env || {}).forEach(env => resolvePluginsPresets(babelrc.env[env]));
 
-  return babelrc;
+  // modify via userland kytConfig's `modifyBabelConfig`
+  try {
+    const mergedBabelrc = config.modifyBabelConfig(babelrc, options);
+    if (config.debug) {
+      logger.debug('Merged babel configuration:', mergedBabelrc);
+    }
+    return mergedBabelrc;
```
tizmagik: This is a misnomer right? There's no actual merging going on.

utils/jest/__tests__/preprocessor.test.js
```diff
@@ -0,0 +1,23 @@
...
+
+const fakeBabelResult = {};
+
+jest.setMock('../../../config/babel', jest.fn(() => fakeBabelResult));
+const babel = require('../../../config/babel');
+
+describe('preprocessor', () => {
+  it('calls the babel utility and babel-jest\'s createTransformer', () => {
+    const preprocessor = require('../preprocessor'); // eslint-disable-line global-require
+    expect(preprocessor).toBe(fakeTransformer);
```
tizmagik: Isn't this just testing Jest's ability to mock?

utils/jest/__tests__/preprocessor.test.js
```diff
@@ -0,0 +1,23 @@
...
+    expect(babel.mock.calls[0][0]).toEqual({
+      type: 'test',
+      environment: 'test',
+    });
+    expect(babelJest.createTransformer).toBeCalledWith(fakeBabelResult);
+  });
+});
```
tizmagik: Thanks for adding this test!! 🙏

config/babel.js
jaredmcdonald: yeah, maybe `modifiedBabelConfig` would be more accurate?

tizmagik: Yea that sounds good

utils/jest/__tests__/preprocessor.test.js
jaredmcdonald: although, wait, I added it to verify that `preprocessor` exports the result of `createTransformer`

tizmagik: Ah ok gotcha!

config/babel.js
jaredmcdonald: @tizmagik - done

delambo:
Can we pause this? I'd like to see how #250 feels first.

jaredmcdonald:
sounds good to me @delambo--I'll close the PR for now, can always reopen

PR #247, opened 2016-10-15
Adds multiple server entry handling

delambo:
- adds support for server webpack configuration with multiple server entries
- adds validation for server entry since we assume that there is at least one server entry named "main"

This pull request does not include any breaking changes to the api.

### DIFF
--- a/utils/__mocks__/webpackCompiler.js
+++ b/utils/__mocks__/webpackCompiler.js
@@ -19,7 +19,19 @@ const multiEntryOptions = {
   devServer: 'devServer',
   output: {
     path: 'realPath',
-    publicPath: 'killPath',
+    publicPath: 'publicPath',
+  },
+};
+
+const stringLiteralEntryOptions = {
+  entry: {
+    main: 'main.js',
+    additional: 'more.js',
+  },
+  devServer: 'devServer',
+  output: {
+    path: 'realPath',
+    publicPath: 'publicPath',
   },
 };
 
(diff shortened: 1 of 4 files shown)


utils/__mocks__/webpackCompiler.js
```diff
@@ -1,13 +1,37 @@
...
     path: 'fakePath',
     publicPath: 'publicPath',
   },
 };
+
+const multiEntryOptions = {
+  entry: {
+    main: ['main.js'],
+    additional: ['more.js'],
+  },
```
tizmagik: Note that an entry can also be a string literal. The test would fail in that case.

delambo: I think this works ok since it only uses the key names.

tizmagik: Ah, I see. Maybe adding the test case with a string literal would be useful though.

tizmagik:
@delambo Does this just add 'watcher' support for multiple entries? Because I think the current kyt on master supports multiple entries fine, no?

delambo:
Yeah, sorry, I should have clarified. kyt/Webpack handles multiple entry points. What this pull request does is makes sure that kyt watches multiple entries during development.

utils/__mocks__/webpackCompiler.js
delambo: Added an additional mock and test for string literals.

tizmagik:
Gotcha, sounds good!

PR #248, opened 2016-10-16
Use --save-dev in Quick Start readme

MiniGod:


tizmagik:
Thanks @MiniGod -- Technically `kyt` is not just a devDependency since `kyt start` is used to start the production server, although that's probably not a good enough reason, and we should probably rethink that a bit.

delambo:
I agree with @tizmagik. Thanks for taking the time, @MiniGod.

PR #255, opened 2016-10-16
Replace babel transform-runtime with polyfill

### LINKED ISSUE #227: Use babel-polyfill instead of transform-runtime for client bundle
The babel transform-runtime plugin is meant more for libs and tools, while the polyfill is meant for production applications. Can we remove the plugin and instead use the polyfill for the client bundle?

https://babeljs.io/docs/usage/polyfill/

> If you are looking for something that won't modify globals to be used in a tool/library, checkout the transform-runtime plugin. This means you won't be able to use the instance methods mentioned above like Array.includes.

dting:
Resolves #227

### DIFF
--- a/docs/kytConfig.md
+++ b/docs/kytConfig.md
@@ -37,6 +37,8 @@ modifyWebpackConfig: (baseConfig, options) => {
 }
 ```
 
+> **NOTE:** [babel-polyfill](https://babeljs.io/docs/usage/polyfill/) is included in the Webpack configs' `entry.main[]` array by default. If you use `modifyWebpackConfig()` to modify `entry.main[]` and you wish to keep using the polyfill, be sure to add the new main entries _after_ the polyfill.
+
 Dev Tip:
 [webpack-merge](https://github.com/survivejs/webpack-merge) is a helpful tool for changing and combining Webpack configs.
 
--- a/.babelrc
+++ b/.babelrc
@@ -1,6 +1,5 @@
 {
   "presets": ["react", ["latest", { "modules": false }]],
-  "plugins": ["transform-runtime"],
   "env": {
     "development": {
       "plugins": ["babel-plugin-transform-react-jsx-source"]
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ kyt's base features include:
 
 * A Node server provided for rendering front-end page requests
 * Client and server hot reloading
-* [ES-Latest stable](https://babeljs.io/docs/plugins/preset-latest/) features using Babel.
+* [ES-Latest stable](https://babeljs.io/docs/plugins/preset-latest/) features using Babel and [babel-polyfill](https://babeljs.io/docs/usage/polyfill/)
 * CSS Module and SASS support
 * Inline SVG support
 * Style and script linter rulesets
--- a/config/__tests__/webpack.test.js
+++ b/config/__tests__/webpack.test.js
@@ -1,0 +1,32 @@
+const devClientConfig = require('../webpack.dev.client');
+const devServerConfig = require('../webpack.dev.server');
+const prodClientConfig = require('../webpack.prod.client');
+const prodServerConfig = require('../webpack.prod.server');
+
+describe('webpack.dev.client', () => {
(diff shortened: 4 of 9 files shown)


docs/kytConfig.md
```diff
@@ -37,6 +37,8 @@ modifyWebpackConfig: (baseConfig, options) => {
 }
 ```
 
+> **NOTE:** The default Webpack configs include [babel-polyfill](https://babeljs.io/docs/usage/polyfill/). If you use `modifyWebpackConfig()` to override entry properties and wish to keep using the polyfill be sure to include it manually.
```
tizmagik: Can we update the wording here to mention that the polyfill is included in `entry.main[]` by default and things can be added to the array but should go _after_ the polyfill.

dting: Sounds good. I've updated the note. Thanks!

docs/kytConfig.md
```diff
@@ -37,6 +37,8 @@ modifyWebpackConfig: (baseConfig, options) => {
 }
 ```
 
+> **NOTE:** [babel-polyfill](https://babeljs.io/docs/usage/polyfill/) is included in the Webpack configs' entry.main arrays by default. If you use `modifyWebpackConfig()` to override the entry properties and wish to keep using the polyfill, the new entries should go after the polyfill.
```
tizmagik: Thanks, this is better, although I want to clarify that new entries can be added as needed but care only need to be taken when/if modifying `entry.main[]` in particular. 

What do you think of something like this?:

```
> **NOTE:** [babel-polyfill](https://babeljs.io/docs/usage/polyfill/) is included in the Webpack configs' `entry.main[]` array by default. If you use `modifyWebpackConfig()` to modify `entry.main[]` and you wish to keep using the polyfill, be sure to add the new main entries _after_ the polyfill.
```

dting: That sounds great. I'll make the change and push it up when I have the chance. 

Also, I need to investigate why the e2e tests are failing for this PR. Permissions denied on the starter kyt repos. I'll take a look tonight. Thanks @tizmagik

ccpricenytimes: Hey @dting I think the problem is because the secret holding the api token for the repo isn't being set since you're PR is from a different repo:

https://docs.travis-ci.com/user/environment-variables/#Defining-Variables-in-Repository-Settings

```Similarly, we do not provide these values to untrusted builds, triggered by pull requests from another repository.```

I'm going to keep looking into this. Can you run the e2e tests locally and let me know if they pass? Thanks!

Comment (dting) [inline docs/kytConfig.md]: Hi @ccpricenytimes. I just tried the e2e tests locally and I get the same errors without access to the test repo (`NYTimes/kyt-starter-test.git`). 

```
  KYT CLI
    ✓ installs kyt (58433ms)
    ✕ sets up a starter-kyt (3379ms)
    ✕ sets up with the correct files (1ms)
    ✕ sets up the package json scripts (2ms)
    ✕ runs the lint command (749ms)
    ✕ runs the lint-style command (1704ms)
    ✓ runs the tests command (592ms)
    ✕ runs the build command (580ms)
    ✓ starts the app (1076ms)
    ✓ dev (549ms)
    ✓ proto (597ms)
```

Comment (dting) [inline docs/kytConfig.md]: Thanks @ccpricenytimes. Just ran the tests and they are passing.

```
 PASS  e2e_tests/tests/cli.test.js (215.179s)
  KYT CLI
    ✓ installs kyt (58219ms)
    ✓ sets up a starter-kyt (84703ms)
    ✓ sets up with the correct files (4ms)
    ✓ sets up the package json scripts (3ms)
    ✓ runs the lint command (4809ms)
    ✓ runs the lint-style command (1838ms)
    ✓ runs the tests command (3956ms)
    ✓ runs the build command (16029ms)
    ✓ starts the app (21881ms)
    ✓ dev (10125ms)
    ✓ proto (8914ms)
```

tizmagik:
@dting would like to merge this in, but there's some conflicts that need to be resolved. Can you merge master into your branch?

dting:
Sure. I can do that right now.

tizmagik:
Thanks @dting !

Issue #257, opened 2016-10-19
Enable css-module jest support


khalidsalomao:
What do you thing about enabling react-css-module support in jest tests as described in [jest documetation](https://facebook.github.io/jest/docs/tutorial-webpack.html)?

Currently I need to customize jest configuration in `kyt.config.js`:

``` js
modifyJestConfig: (baseConfig) => {
    // solution for jest / enzime and css-module problem:
    baseConfig.moduleNameMapper['^[./a-zA-Z0-9$_-]+\\.(css|scss)$'] = 'identity-obj-proxy';
    return baseConfig;
}
```

And add the necessary dependencies:
`npm install identity-obj-proxy --save-dev`

tizmagik:
Yes, a PR for this would be great! Currently this just lives as a "TODO" in the code (https://github.com/NYTimes/kyt/blob/master/utils/jest/style.stub.js#L1), but thank you for creating a proper issue @khalidsalomao ! 😄

khalidsalomao:
@tizmagik :+1: 
I will take a look into it!

delambo:
@khalidsalomao - Did you get a chance to check this out? I ran into this problem the other day when I tried to test against a class in an internal project. I'd really like to get a fix for this.

khalidsalomao:
@delambo sorry I didn't check this out yet! I will look into it tomorrow!
As a temporary solution you could simply enable it in the `kyt.config.js` as described above.

delambo:
@ccpricenytimes got this to work in a 0.4.0rc. It should land soon in a final 0.4.0 version.

PR #258, opened 2016-10-19
Enable css-module jest support

khalidsalomao:
Enable css-module support in jest.

Issue: https://github.com/NYTimes/kyt/issues/257
For more details see: https://facebook.github.io/jest/docs/tutorial-webpack.html

### DIFF
--- a/utils/jest/style.stub.js
+++ b/utils/jest/style.stub.js
@@ -1,3 +1,3 @@
-// TODO: use https://github.com/keyanzhang/identity-obj-proxy
+// see: https://facebook.github.io/jest/docs/tutorial-webpack.html
 
-module.exports = {};
+module.exports = 'identity-obj-proxy';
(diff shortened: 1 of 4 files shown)


utils/jest/style.stub.js
```diff
@@ -1,3 +1,3 @@
-// TODO: use https://github.com/keyanzhang/identity-obj-proxy
+// see: https://facebook.github.io/jest/docs/tutorial-webpack.html
 
-module.exports = {};
+module.exports = 'identity-obj-proxy';
```
tizmagik: I don't think this would work, right? It would just export a string literal. You'd have to either import and re-export the module here (messy) or change the mapping in moduleNameMapper in jest.js

How did you test this?

khalidsalomao: My bad! I created a sample project but kept a modifed kyt.config.js.
Let me review this!

khalidsalomao:
@tizmagik strange, I have it working with sass files. I will have to take a deeper look into it!

ccpricenytimes:
I was finally able to get this working. I created a new PR #367  because of all the folder changes moving to a monorepo. Thanks so much for your help @khalidsalomao !

PR #270, opened 2016-10-24
Adds yarn support with npm fallback

### LINKED ISSUE #254: Support Yarn
https://github.com/yarnpkg/yarn

Two steps:
- [ ]  Test to make sure kyt works successfully when installed with yarn
- [ ] add yarn support for starter-kyts in setup. Should still support npm.

tizmagik:
Closes #254


elisechant:
just why? have one or the other, won't this cause more possible dependency issues that its solves?

tizmagik:
@elisechant this shouldn't cause any issues I'm aware of, do you have a specific case in mind?

elisechant:
@tizmagik yarn's dependency tree is more strict than npm's without a shrink-wrap. Potentially you could have developers with environments that have dependency discrepancies. I think the solution here should look like either using npm (perhaps with shrink-wrap) or yarn.

tizmagik:
Thanks for the clarification @elisechant 

The main benefit of this PR is to speed up CI, especially the e2e tests. Users are free to continue using npm, even if yarn is installed globally on their machine.

The only impact on users is for the `setup` command where the dependencies are prescribed in the starter-kyts. If a custom starter-kyt ends up having issues with yarn then we can look into adding an option to force kyt to use npm. I'm happy to look into that. Are you aware of any starter-kyts which you know would have an issue with yarn?

With all that being said, the reason why yarn is so compelling is not only because it's faster, but also because it's precisely more strict than npm+shrinkwrap. Shrinkwrap is unfortunately notoriously inconsistent and difficult to work with, so having the option of using yarn (iff it's installed globally) I see as a real benefit. 

I'm sure we'd also be happy to accept a PR that adds an option to force using npm on `setup`, I don't see a reason why not.

cli/actions/setup.js
```diff
@@ -19,6 +20,7 @@ const {
 const kytPkg = require(path.join(__dirname, '../../package.json'));
 
 module.exports = (config, flags, args) => {
+  logger.info('kyt version: ', kytPkg.version);
```
ccpricenytimes: what's this for?

package.json
```diff
@@ -7,7 +7,7 @@
     "kyt": "cli/index.js"
   },
   "engines": {
-    "node": ">=6.1.0"
+    "node": ">=6.0.0"
```
ccpricenytimes: whats the reasoning for this?

tizmagik:
This is technically blocked by `yarn v0.18` being released.

The issue is yarn resolves relative paths incorrectly. Fixed in https://github.com/yarnpkg/yarn/pull/1498

**Possible Workarounds:**
1. Resolve relative paths before building them up in the e2e test, but I'd rather leave them as is.
2. Keep using `npm` for the e2e tests, this will require specifying which package manager to use on `setup` which I'm looking into adding now anyway.

sean-clayton:
v0.18.0 has been released (v0.18.1 is the latest as of this comment):
https://github.com/yarnpkg/yarn/releases/tag/v0.18.0

scripts/yarn-install-modules.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
```
delambo: Instead of creating an optional yarn-install-modules file, do you all think it would be better to add a conditional to the install-modules file that will use yarn if it exists. Something like:

```
if ! type "yarn" > /dev/null; then
  yarn 2>&1
else
  npm i 2>&1
fi
```
I use bootstrap.sh a lot to update, so it would be helpful if the install-modules script used yarn.

PR #281, opened 2016-10-27
Update babel.js

vishnubhagwan:
# 134


ccpricenytimes:
Hi @vishnubhagwan, closing this for now as we would like to keep that note in the docs only. thanks

Issue #288, opened 2016-10-27
Replace default npm init test script on setup


delambo:
On `npm init` a "test" script is added in the package.json scripts section. It looks like the following:

```
"test": "echo \"Error: no test specified\" && exit 1"
```

We should sniff this out [when we add scripts](https://github.com/NYTimes/kyt/blob/master/cli/actions/setup.js#L68) during our `setup` command and overwrite the script with our kyt test script: `kyt test`.

aterranova-bv:
I'll tackle this one, happy hacktoberfest!  👻 🎃 👻 🎃

jaredmcdonald:
addressed in #293, closing

PR #293, opened 2016-10-27
Overwrites default npm init test script on setup

aterranova-bv:
# Issue

Addresses #288.
# Details

Previously, if a `test` script was found in `package.json`, `setup` would keep the original `test` script and create an additional `kyt:test` script that would run `kyt test`.

Now, `setup` checks to see if the existing `test` script is simply the `npm init` default of `echo \"Error: no test specified\" && exit 1`, and if so, overwrites `test` itself to be `kyt test`.  Note that the additional `kyt:test` script is no longer created in this case.
# Notes

Unit testing this code change proved difficult because it lives in a function internal to `setup`. There are currently no `setup` unit tests, and altering the [pkg.json](https://github.com/NYTimes/kyt/blob/master/e2e_tests/pkg.json) that the end-to-end tests load wasn't advisable.

As such, I have only manually verified these changes but the test cases would be:

| Scenario | Expected Behavior |
| --- | --- |
| `package.json` has `npm init` default test script | `test` script becomes `test: kyt test` |
| `package.json` has `test` script that is not the `npm init` default | `test` script is unaltered, and a new `kyt:test: kyt test` entry is added |
| `package.json` has no `test` script | `test: kyt test` script is added |

Happy hacktoberfest! 👻 🎃 👻 🎃

### DIFF
--- a/cli/actions/setup.js
+++ b/cli/actions/setup.js
@@ -87,7 +87,7 @@ module.exports = (config, flags, args) => {
     }
 
     // This is the default test script added by 'npm init'.
-    const npmInitDefaultTestScript = 'echo \"Error: no test specified\" && exit 1'; // eslint-disable-line no-useless-escape
+    const npmInitDefaultTestScript = 'echo "Error: no test specified" && exit 1';
 
     commands.forEach((command) => {
       let commandName = command;
--- a/cli/actions/setup.js
+++ b/cli/actions/setup.js
@@ -86,6 +86,9 @@ module.exports = (config, flags, args) => {
       commands = uniq(commands.concat(tempScripts));
     }
 
+    // This is the default test script added by 'npm init'.
+    const npmInitDefaultTestScript = 'echo \"Error: no test specified\" && exit 1'; // eslint-disable-line no-useless-escape
+
     commands.forEach((command) => {
       let commandName = command;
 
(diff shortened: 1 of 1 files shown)


cli/actions/setup.js
```diff
@@ -86,6 +86,9 @@ module.exports = (config, flags, args) => {
       commands = uniq(commands.concat(tempScripts));
     }
 
+    // This is the default test script added by 'npm init'.
+    const npmInitDefaultTestScript = 'echo \"Error: no test specified\" && exit 1'; // eslint-disable-line no-useless-escape
```
jaredmcdonald: I think eslint considers it a "useless escape" because the escape character needs to be escaped, i.e., this should be `'echo \\"Error...'`. could you write an end to end test (in `e2e_tests/tests`) to verify?

jaredmcdonald: (or, alternately, it just ignores the escape character, in which case it's ok to omit the escapes from `npmInitDefaultTestScript` and remove the eslint-disable)

aterranova-bv: Hi @jaredmcdonald, the `npm init` default test script has the escape characters in it, and I wanted to check for that string exactly.  I think this may be a valid use case for ignoring the lint rule on this line.

jaredmcdonald: @aterranova-bv it escapes it for json--when you read it into node it will just be a string without those escapes.

```
test 🙀  > cat package.json
{
  "name": "test",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "",
  "license": "ISC"
}
test 🙀  > node
> const pkg = require('./package.json')
undefined
> pkg.scripts.test
'echo "Error: no test specified" && exit 1'
```

jaredmcdonald: also:

```js
'echo "Error: no test specified" && exit 1' === 'echo \"Error: no test specified\" && exit 1'
true
```

aterranova-bv: Yep, you're totally right.

aterranova-bv:
PR updated to omit escape characters and remove eslint line disable directive.

Issue #292, opened 2016-10-27
Test the docs!


ccpricenytimes:
Try to setup kyt using only the README instructions and documentation. File issues for things that are unclear, or PRs adding clarification!

Dog:
I'm going through this issue now.

PR #294, opened 2016-10-27
README.MD: Updated Quick Start Section

Dog:
#292

I didn't really have any problems. I just tried to make it a bit more explicit. It seemed to be assumed someone would know to run `npm init` while all the other commands were given.

I changed the markdown numbering so it keeps track of the number of steps.

Let me know if anything else is needed.


ccpricenytimes:
Hi @Dog thanks for this. 

We're currently working on a more long form tutorial and we'd prefer to keep the quick start as concise as possible. Would you be interested in reading the tutorial when we have a draft? 

Closing this for now.

Issue #284, opened 2016-10-27
Update postcss-loader


ccpricenytimes:
postcss-loader should be upgraded to 1.0.0

Should be tested against starter-kyts to make sure everything still works. (Bonus points for a new e2e test!)

ccpricenytimes:
resolved by #295

PR #295, opened 2016-10-27
Update postcss-loader

potatogopher:
#284


ccpricenytimes:
Thanks @potatogopher !

PR #297, opened 2016-10-28
updated step 2 of quickstart

steve-chen-nyc:
Hey here at Hacktoberfest! Had a quick idea for simple contribution. 

I think it would be a little more clear esp if you are new to Node to run NPM Init to create the package.json file vs just asking users to create a package.json file.


ccpricenytimes:
Thanks @steve-chen-nyc

Issue #179, opened 2016-09-23
babel-transform-react-jsx-source


tizmagik:
I didn't know this transform existed, but it seems super useful. It basically gives you a better idea of where the react proptype warnings are coming from by giving specific files and line number references in the error. 

https://twitter.com/dan_abramov/status/779308833399332864

We should be able to add in this transform in 'dev' mode pretty easily.

PR #301, opened 2016-10-29
Add transform-react-jsx-source webpack plugin

donnanicolas:
This attempts to solve #179. I don't know if this what you need.

All the test run and pass, let me know if anything else is needed

Cheers


delambo:
Thanks for looking into this @donnanicolas! We have another pull request that takes care of this by using a babelrc plugin instead. I'm going to close this one in favor of that. Thanks, again.

PR #302, opened 2016-10-29
Add babel-plugin-transform-react-jsx-source to dev (#179)

### LINKED ISSUE #179: babel-transform-react-jsx-source

zdoc01:
Resolves #179. Added the new plugin definition to the root `.babelrc` to follow standard conventions. I tested the changes against the static starter-kyt by changing some of the prop requirements on one of the generated components. Here's a few screenshots of the before and after:

Before
screen shot 2016-10-29 at 12 42 51 pm

After
screen shot 2016-10-29 at 12 42 15 pm

Also - I noticed the `.babelrc` doesn't support defining plugins and presets without the `babel-*` prefix because of the way they are being resolved in `config/babel.js`. I think I have a fix for this though, so I'm going to create an issue and open a separate PR.

Let me know if you have any questions!

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -3,7 +3,7 @@
   "plugins": ["babel-plugin-transform-runtime"],
   "env": {
     "development": {
-      "plugins": []
+      "plugins": ["babel-plugin-transform-react-jsx-source"]
     },
     "production": {
       "plugins": [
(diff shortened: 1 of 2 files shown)


delambo:
This looks great. I'd love to get @tizmagik's eyes on this one just to confirm that this fully covers #179.

tizmagik:
This is great, thanks @zdoc01 !

zdoc01:
No problem!

PR #304, opened 2016-10-29
Clarify docs

MrClement:
#292 

The docs worked quite well. A few quick things:

- Make it clearer that `node_modules/.bin/kyt setup` is how you run the
  setup command for the first time.
- In docs/commands, make the list of commands at the top links
- Update the kyt.config.js example in `proto` to use the correct key


ccpricenytimes:
Thanks @MrClement !

PR #305, opened 2016-10-29
Support plugin/preset definitions without prefixes in .babelrc (#303)

### LINKED ISSUE #303: Support plugin/preset definitions without prefixes in .babelrc
We currently cannot specify babel plugins and presets without the `babel-*` prefix in `config/babel.js` because of the way they are resolved. This is a standard convention that's highlighted in the babel docs for a given plugin or preset (see [these babel-preset-react docs](https://babeljs.io/docs/plugins/preset-react/#usage) as an example).

zdoc01:
Resolves #303. Still supports `babel-*` plugins and presets in `.babelrc`, as well as without the prefixes. Also still supports relative and absolute paths.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,19 +1,19 @@
 {
-  "presets": ["babel-preset-react", ["babel-preset-latest", { "modules": false }]],
-  "plugins": ["babel-plugin-transform-runtime"],
+  "presets": ["react", ["latest", { "modules": false }]],
+  "plugins": ["transform-runtime"],
   "env": {
     "development": {
       "plugins": []
     },
     "production": {
       "plugins": [
-        "babel-plugin-transform-react-remove-prop-types",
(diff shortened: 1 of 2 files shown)


delambo:
@zdoc01 - Can you quickly look into making a unit test for this. I think we stub out the babelrc so this may be hard to test. Since this is pretty simple and will only temporarily live in the project I'm not terribly concerned about a test but if you find that it is easy to add, please do. Also, feel free to do it in a follow-up pull request.

delambo:
Closed since I know you're working against the clock @zdoc01. Please see my comment above and followup with a pull if it's possible.

zdoc01:
@delambo - Sorry, I didn't get a chance to respond yesterday. I'll happily take a look at adding a test when I get the chance. May not be today, but hopefully tomorrow at the latest.

PR #314, opened 2016-11-01
document possible kyt setup repository url formats [ci skip]

### LINKED ISSUE #313: add note about ssh key to setup docs
we've seen a bunch of issues filed that boil down to not having an ssh key set up for `kyt setup` with a ssh repository (#308, #268); let's add a quick note to the `setup` docs clarifying this (probably linking to the GH guide on setting up an ssh key) and note that you can use an https url

jaredmcdonald:
fixes #313 

I also changed the example commands to set up the recommended starter kyts to https urls (it's authless because they're public, so probably easier overall)

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -38,11 +38,16 @@ If kyt finds any files with duplicate names, it will back up your file before re
 
 Running `kyt setup` will give you the option to install the [default starter-kyt.](https://github.com/NYTimes/kyt-starter-universal)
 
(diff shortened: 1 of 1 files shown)


ccpricenytimes:
We should wait to merge this until the https change goes out yeah?

jaredmcdonald:
@ccpricenytimes - nope, should be good to go whenever, this just documents existing behavior 👍

PR #316, opened 2016-11-02
Adds rfc template

delambo:
We're testing a new rfc process in github Issues. So far the idea is to copy and use the rfc template from the Rust repository. We plan on working out the details for the process - how are rfcs labeled, what are their various stages, how are they accepted - as we roll out the first couple proposals.

Feel free to leave ideas here about the initial rfc process. I'll open an Issue so that we can further discuss the final process.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,6 +18,10 @@ We recommend forking kyt and creating a test project with a starter-kyt for loca
 ### Testing kyt
 Instructions TK
 
+## Create an RFC
+
+If you want to propose a large feature idea or architecture change you should consider submitting an RFC. It's often helpful to get feedback on your concept in an issue before starting the RFC. RFCs are an evolving process in the kyt repository so expect a lot of changes and guidelines in the future. You can find the kyt RFC template [here](/rfc/template.md).
+
 ## Build a starter-kyt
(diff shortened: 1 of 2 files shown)


tizmagik:
Consider [issue templates](https://help.github.com/articles/creating-an-issue-template-for-your-repository/)

delambo:
I'm not exactly sure how issue templates work, but from what I gather the text in the ISSUE_TEMPLATE is inserted into every new issue. If that's right, I'd prefer to leave it in a standalone file for now since I'd like to reserve rfcs for big stuff.

tizmagik:
Right, but you don't have to insert the entire RFC template in the issue template, you can just point folks to it during issue creation behind HTML comment tags (so they don't show up in the actual issue). So, for example, something like this:

```
<!--
    Writing an RFC?: Please use the template in rfc/template.md

    Writing a bug report/feature request?: Please fill in the following fields:
-->    

### Versions of kyt, node, OS

### Bug or Feature?

### Current Behavior

### Desired Behavior
```

And so on...

delambo:
Ah, got ya.

delambo:
Got the issue template in: https://github.com/NYTimes/kyt/pull/317

PR #318, opened 2016-11-03
Update unit tests for changes made to resolve #303

### LINKED ISSUE #303: Support plugin/preset definitions without prefixes in .babelrc

zdoc01:
See issue #303 and PR #305 for more information. Covers cases in `configs/babel.js` where babel dependencies are resolved. Let me know if you have any issues with the approach.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -2,7 +2,7 @@
   "presets": ["react", ["latest", { "modules": false }]],
   "env": {
     "development": {
-      "plugins": ["babel-plugin-transform-react-jsx-source"]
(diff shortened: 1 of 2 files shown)


delambo:
Thanks @zdoc01!

PR #336, opened 2016-12-10
Run test from root directory

delambo:
This adds the ability to test all of the packages from the root directory, removing the need for local packages to manage jest dependencies and scripts. ~~Since it needs to be there, we reference the jest dependency in the kyt-core package.~~ We'll still need to manage a separate jest dependency in kyt-core for user projects.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,6 @@
   "dependencies": {
     "ps-tree": "1.1.0",
     "shelljs": "0.7.5",
-    "jest": "17.0.3"
+    "jest": "16.0.1"
   }
 }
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -22,8 +30,8 @@
   },
   "homepage": "https://github.com/NYTimes/kyt#readme",
   "dependencies": {
-    "jest": "16.0.2",
     "ps-tree": "1.1.0",
-    "shelljs": "0.7.5"
+    "shelljs": "0.7.5",
+    "jest": "17.0.3"
```
ccpricenytimes: should we upgrade Jest to 17 in kyt-core too? There may be some breaking changes we'll want to look into. How much do we care about keeping them in sync?

delambo: Let's not right now. I think we should keep them in sync. Fixed.

PR #338, opened 2016-12-10
0.4.0 docs

ccpricenytimes:
Add any other documentation for 0.4.0 changes should go in this branch.


ccpricenytimes:
@delambo should we move the old linter style docs into their respective package dirs?

PR #341, opened 2016-12-10
Adds bootstrap scripts

delambo:
Most of us have been on the fence about using [Lerna](https://github.com/lerna/lerna) to manage our monorepo. Until we decide, I thought we could use a bootstrap script for setting up the project and managing node modules across the packages.

Notes:

1) I intentionally left out the starter kyts because I feel like less work is done there and can easily be handled manually.
2) I remove node_modules directories in the background/parallel, but chose to install the packages sequentially because the npm output gets messy.

### DIFF
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -1,0 +1,6 @@
+#!/bin/sh
+
+source scripts/remove-modules.sh
+source scripts/install-modules.sh
+
+cd packages/kyt-cli && npm link && echo "👍  linked kyt-cli"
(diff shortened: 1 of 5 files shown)


scripts/get-packages.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+PACKAGES=( "" )
+ROOT=`pwd`
+
+for FILE in packages/*; do
```
ccpricenytimes: should we call this variable DIR since its getting directories?

scripts/bootstrap.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
```
ccpricenytimes: silly, but can we add 🔥 and ✅ ? haha

PR #347, opened 2016-12-13
support user .babelrc

### LINKED ISSUE #322: RFC: respect user .babelrc
## Summary
[summary]: #summary

kyt should respect user-defined `.babelrc` files for Babel configuration, and create Babel presets (for use in starter-kyts, and as the kyt default) to encapsulate opinions. (This is a longer exploration of #250.)

A proof-of-concept branch is available [here](https://github.com/NYTimes/kyt/tree/babelrc) (note, this is a work in progress).

## Motivation
[motivation]: #motivation

This change would dramatically simplify user and starter-kyt Babel configuration, providing an easy (and canonical) way to extend default configuration.

The current mechanism for modifying Babel configuration in kyt.config's `modifyWebpackConfig` is a bit unwieldy--users have to look up the Babel loader in the list of Webpack loaders and modify its `options`. Jest also uses Babel to preprocess files, so we currently have a bit of a [hack](https://github.com/NYTimes/kyt/blob/master/utils/jest/preprocessor.js#L9-L13) in place where we construct the Webpack config in order to pull out the Babel config for Jest.

Essentially we'd go from user projects looking like this:

```js
// kyt.config.js
modifyWebpackConfig(config, options) {
  const babelLoader = config.loaders.find(loader => loader.loader === 'babel-loader')
  babelLoader.options.plugins.push(myPlugin)
  return config
}
  ```
  
to this:

```json
// .babelrc
{
  "presets": ["kyt-core"],
  "plugins": ["my-plugin"]
}
```
  
The new format has a couple of benefits:

1. it doesn't require knowledge of Webpack, or its internal structures

1. configuration is more "declarative"

1. it's canonical Babel usage; compare this with creating a kyt-specific abstraction that still requires knowledge of Babel (see Alternatives (1) and (2))

1. kyt codebase gets simpler; no longer do we have to support the hack outlined above to support Jest, for instance

## Proposed design
[design]: #proposed-design

To begin with, create separately packaged [babel presets](https://babeljs.io/docs/plugins/#creating-a-preset); we'd need one for kyt core opinions (essentially just extending [preset-latest](http://babeljs.io/docs/plugins/preset-latest/), and then we'd create additional presets per framework, each extending the kyt core preset.

kyt will specify the core Babel preset as a dependency and use it as a default for when no user `.babelrc` is found; otherwise, kyt will respect the `.babelrc` in the root of the user project. For Webpack (`kyt build`, `kyt dev`, etc.), this isn't any additional work (we'd just need to remove the line in the loader that ignores `.babelrc`); for Jest (`kyt test`), we'd need to resolve each plugin and preset relative to the user project root.

starter-kyts could then optionally include a `.babelrc` and presets (as dependencies) that get copied over during `kyt setup`.

## Drawbacks
[drawbacks]: #drawbacks

1. Overriding or disabling preset defaults is harder than with a callback interface (see Alternatives (1)). We could mostly work around this with [preset options](https://babeljs.io/docs/plugins/#pluginpresets-options), but we'd have to manage those options closely to avoid sprawl. In practice this might not be too terrible.

1. Users would have to manage more dependencies: they would have to bump Babel preset versions in addition to kyt. However, the presets and kyt core would likely be fairly decoupled.

1. There's no way to distinguish between server and client code in a `.babelrc`, so development tools like React Hot Loader (which we don't want on the server) will still have to be injected into the options for the Babel loader via Webpack configuration

## Alternatives
[alternatives]: #alternatives

1. Provide a callback interface in `kyt.config.js`, `modifyBabelConfig`, which allows users to make whatever modifications they'd like to the base configuration, returning the new config for use in the Webpack loader and in the Jest preprocessor. See #134.

1. Abstract away Babel configuration entirely. This doesn't seem like a viable option--teams might need to, for example, add an application-specific [Babel plugin for Relay queries](https://facebook.github.io/relay/docs/guides-babel-plugin.html), or support various early-stage syntax. We'd either end up reimplementing a `.babelrc`-like format to define additional plugins and presets, or handicapping users, neither of which is desirable.

jaredmcdonald:
fixes #322


packages/kyt-core/config/webpack.base.js
```diff
@@ -5,75 +5,104 @@
...
-module.exports = options => ({
-  node: {
-    __dirname: true,
-    __filename: true,
-  },
+module.exports = (options) => {
+  const hasBabelrc = shell.test('-f', userBabelrcPath);
+  if (!hasBabelrc) {
+    logger.warn('No user .babelrc found. Using kyt default babel preset...');
+  }
```
ccpricenytimes: Just checking these are the only changes you made in this file yah? The rest of the diff just looks like spacing

packages/babel-presets/babel-preset-kyt-core/.eslintrc.json
```diff
@@ -0,0 +1,5 @@
+{
+  "rules": {
+    "no-var": 0
```
tizmagik: How come?

packages/kyt-cli/cli/actions/setup.js
```diff
@@ -282,6 +283,17 @@ module.exports = (flags, args) => {
     logger.task('Created .editorconfig file');
   };
 
+  const createBabelrc = () => {
+    // back up existing .babelrc, if it exists
+    if (shell.test('-f', userBabelrcPath)) {
+      const mvTo = path.join(userRootPath, `.babelrc-${date}.bak`);
```
tizmagik: Maybe append a time stamp so there's no collisions

packages/kyt-core/utils/jest/preprocessor.js
```diff
@@ -1,17 +1,57 @@
...
+ * @param  {String} prefix One of either 'babel-plugin' or 'babel-preset'
+ * @param  {String} dep    The dependency
+ * @return {String}        The normalized dependency
+ */
+const normalizeDep = (prefix, dep) => {
+  let resolved;
+  if (dep.indexOf('babel') !== 0) {
+    try {
+      resolved = resolve.sync(`${prefix}-${dep}`, { basedir: userRootPath });
+    } catch (e) { /* eslint-disable no-empty */ }
```
tizmagik: What does this error case mean?

packages/babel-presets/babel-preset-kyt-core/.eslintrc.json
delambo: I'm also curious of the argument for using < ES2015 in the presets.

packages/kyt-core/config/webpack.base.js
jaredmcdonald: yeah, I had to indent everything because it just immediately returned an object before

packages/babel-presets/babel-preset-kyt-core/.eslintrc.json
jaredmcdonald: maybe this is naive but I figured there's nothing about these presets that really requires them to be used inside kyt, so might as well make them es5 compatible

packages/kyt-core/utils/jest/preprocessor.js
jaredmcdonald: `resolve.sync` throws if you try to resolve something that doesn't exist. maybe we should log the error?

packages/kyt-cli/cli/actions/setup.js
jaredmcdonald: `date` is set to `Date.now()` above so I think we should be ok here?

tizmagik: Oh sorry missed the `date` bit, yeap perfect 👍

jaredmcdonald:
@ccpricenytimes agreed that this needs docs--should I do it here, or in the 0.4.0 docs branch?

packages/kyt-core/utils/jest/preprocessor.js
tizmagik: I think so right? Because this would be an exceptional case (where there's a babel plugin referenced that doesn't exist so it should fail loudly, I think.

packages/babel-presets/babel-preset-kyt-core/.eslintrc.json
jaredmcdonald: FWIW these `.eslintrc.json`s aren't even honored right now since the root package's `lint` ignores them (to work around needing to get the eslint config cascade working)

packages/kyt-core/utils/jest/preprocessor.js
jaredmcdonald: @tizmagik added a commit to be noisier about this, f6ad967

PR #351, opened 2016-12-16
fixes top level lint command

delambo:
Here's my idea: We should configure one top-level lint command that lints against all of the packages and that references our kyt linter configuration. That should simplify the monorepo and give us a chance to do some dog wolfing of our own linter configuration.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
 language: node_js
 node_js:
   - 6.0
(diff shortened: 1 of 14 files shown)


.travis.yml
```diff
@@ -1,4 +1,4 @@
 language: node_js
 node_js:
   - 6.0
-script: npm install && cd packages/kyt-core && npm install && cd ../kyt-cli && npm install && cd ../kyt-utils && npm install && cd ../.. && npm run lint && npm test && npm run e2e
+script: npm run update && npm run lint && npm test && npm run e2e
```
ccpricenytimes: so much better haha

PR #354, opened 2016-12-20
remove deprecated ESLint rule `prefer-reflect`

pishposh:
Not sure why we enabled this rule in the first place, but it’s been deemed [pointless](http://eslint.org/docs/rules/prefer-reflect) by ESLint’s maintainers. (I haven’t updated ESLint or any other dependencies in this p.r., out of caution, having shaved enough yak tonight.)


pishposh:
Whoops, scratch that.

Issue #356, opened 2016-12-21
Feature: Use babel-preset-node6 / node7 for server bundle


c0b:
since the tool [need NodeJS (v6.0+ required)](https://tom29739.github.io/kyt/#quickstart), I think it makes more sense to build server bundle to babel-preset-node6 or node7, and may give an cli switch to output es5 in server bundle if the user really want that;

Refer https://github.com/NYTimes/kyt/issues/216#issue-181175587

right now my workaround is to use a `modifyWebpackConfig` recipe, but if the option can be default that would be better

```js
modifyWebpackConfig: (baseConfig, options) => {

  if (options.type === 'server') {
    const babelLoader = baseConfig.module.rules.find(
          loader => loader.loader === 'babel-loader');
    babelLoader.options.presets = ['node6', 'react'];
  }

  return baseConfig;
}
```

sean-clayton:
Wouldn't it be best that it goes inside your own `.babelrc` file? No need to modify your webpack config, either.

tizmagik:
Thanks @c0b this actually a good point and we've been discussing this internally for some time. To take this a step further, we're planning on adding support for `babel-preset-env` in #359 -- that, along with kyt 0.4.0 release, which will make babel configuration easier should cover all use cases very cleanly. So I'm going to close this issue for now, but feel free to discuss in those other PRs/issues, or post back here if I missed something. Thanks!

c0b:
@sean-clayton is there a way to use .babelrc to change server build only? please consider add an example in project page or the https://github.com/NYTimes/kyt/tree/master/docs ?
for client build I think it's still preferred to transpile to ES5 for better browser coverage

PR #359, opened 2017-01-07
babel-preset-latest => babel-preset-env

### LINKED ISSUE #312: Explore babel-preset-env
With #255 we now have `babel-polyfill` in place both on the client and server for necessary ES6+ support. In the case of the client, relatively modern browsers don't need the entire polyfill, and for the server since node 6 is already a pre-req for kyt, most of the polyfill is also unnecessary there.

Although it's arguably still early for [babel-preset-env](https://github.com/babel/babel-preset-env), I think it might be good to experiment with it and see if we can use it to polyfill only what's necessary for newer browsers in the client bundle and node 6+ in the server bundle.

tizmagik:
Closes #312 

Still todo for this PR:

- [ ] should use `"node":"current"` babel target setting for server build ([docs](https://github.com/babel/babel-preset-env#support-a-target-option-node-current-to-compile-for-the-currently-running-node-version))
- [ ] add note about browser targets for userland babel configs

cc @jaredmcdonald


c0b:
coming from #356 I'm not sure how to use this PR if I merge it locally; as the docs mentioned I have created a .babelrc but it seems kyt is not honoring it

tizmagik:
Ah, sorry for the confusion. The latest kyt on npm (0.3.0) does not honor a local .babelrc -- that will happen in 0.4.0 (which is what master reflects currently). For now, you'll need to manually override babel settings in modifyWebpackConfig() in kyt.config.js

delambo:
@tizmagik - Would it be safe to merge this as-is and followup with the other tasks?

tizmagik:
Yea maybe, let me see what's going on with the tests

tizmagik:
Closing in favor of #531

PR #377, opened 2017-02-03
Adds monorepo bootstrap

delambo:
- Moves all packages to the `/packages` directory
- Changes the bootstrap script from ugly bash code to some decent looking JavaScript which should mostly work cross-platform
- symlinks local packages across the monorepo package dependencies


bootstrap.js
```diff
@@ -0,0 +1,58 @@
...
+
+// Install the root package.json first so that we can use shelljs.
+installPackage(process.cwd());
+const shell = require('shelljs');
+
+// Install all of the monorepo packages.
+packages.forEach(pkg => installPackage(pkg.path));
+
+// Symlink monorepo package dependencies to local packages.
+packages.forEach((pkg) => {
```
elisechant: how does this system resolve dependency clashes when you flatten here? for example: if one of your monorepo packages depends on [given module]@2.2.0 and a sibling package depends on version [given module]@1.9.0. Do you assume that your project will include both versions of dependency?

delambo: All of the  kyt-specific dependencies in the monorepo should be in sync with the latest published versions. Symlinking makes it easy to do local development on a specific kyt package and its respective kyt dependencies (eg changes to kyt-core and kyt-utils). It's true that this makes the assumption that things will stay in sync, but the symlinks are also easy to override after bootstrap if there is an issue or you want to test against a specific version. Let me know if I can clarify more.

elisechant: so the assumption is that the last package which contains dependency of said version will be the chosen version as it was the last symlinked? Wondering if this has caused or might cause interdependency related errors in the past/future? Do you have a recommended approach to ensuring the dependencies as you say as "latest published versions"? do you suggest to use `@latest` for example? Or how do you sanity check this in terms of upgrading workflow?  Should this code emit warnings to handle this case for example?

ccpricenytimes: This tool is just for local development so you can make changes without having to constantly reinstall node_modules. All published packages will have pinned dependencies.

For example:
I want to add a new utility to kyt-utils to use in a command in kyt-core. When I bootstrap for local development kyt-utils is symlinked in my node_modules for kyt-core and I can make changes directly in my files and see those changes as I run and test kyt-core locally. When I'm ready to commit changes I would update and publish a new version of kyt-utils and then use that pinned version as the dependency of kyt-core. 

Dependencies that we don't manage (eg. jest, shelljs) will always use versions from npm even in local development.

CONTRIBUTING.md
```diff
@@ -24,23 +24,29 @@ We feature [recommended starter-kyts](/docs/commands.md#recommended-starter-kyts
...
+```
+git clone git@github.com:NYTimes/kyt.git
+yarn run bootstrap
+```
+There are commands for bootstrapping, testing and linting all of the monorepo packages in the root directory package.json file. 
 
 ### bootstrap
 
-Bootstrap` will set you up with a clean slate. Every time it is run, it will remove and re-install the node_modules across all of the kyt packages. It will also link `kyt-cli` so you can run it locally on the command line.
+Bootstrap` will set you up with a clean slate. Every time it is run, it will remove and re-install the node_modules across all of the packages, npm link `kyt-cli` and `kyt` so you can run them locally on the command line, and symlink local monorepo dependencies..
```
ccpricenytimes: I think you need to move the `

packages/kyt-starter-static/build/publicAssets.json
```diff
@@ -0,0 +1 @@
+{"main":{"js":"http://localhost:3001/main.js"}}
```
ccpricenytimes: we should add these files to the gitignore

packages/kyt-starter-universal/package.json
```diff
@@ -24,6 +24,7 @@
   },
   "devDependencies": {
     "enzyme": "^2.4.1",
+    "kyt": "next",
```
ccpricenytimes: can we double check to make sure this works during setup

delambo: I think we should pin it to the latest published version of kyt. Updated.

packages/kyt-starter-static/package.json
```diff
@@ -23,6 +23,7 @@
   },
   "devDependencies": {
     "enzyme": "^2.4.1",
+    "kyt": "0.3.0",
```
ccpricenytimes: this should be 0.4.0 since these have .babelrc's yeah?

PR #400, opened 2017-02-15
Expand version range of yarn for Travis

julianvmodesto:
### Issue
[Yarn on Travis](https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#Travis-CI-supports-yarn) is installed as the latest, which is now 0.20.3. Our bootstrap.js file checks for 0.19.0, so builds fail with:

```
0.20.3
❌  update your version of yarn: npm i yarn@^0.19.0 -g
```

I have >= 0.19.0 for now unless there's a tighter version range we want.

### DIFF
--- a/bootstrap.js
+++ b/bootstrap.js
@@ -34,7 +34,7 @@ const shell = require('shelljs');
 const semver = require('semver');
 
 // Make sure that we're all using the same version of yarn.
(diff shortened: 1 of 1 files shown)


delambo:
Maybe you should bump it to 20, while you're at it

Issue #397, opened 2017-02-14
NodeJS source-map using source-map-support


bkniffler:
Hey, 
currently source-maps of server main.js are not being used by nodejs. Maybe you guys could try and use https://www.npmjs.com/package/source-map-support to fix that?

bkniffler:
Btw. this can easily be done by adding the following plugin to server config (since it needs to be included in the server entry point):

`new webpack.BannerPlugin('require("source-map-support").install();', {raw: true,entryOnly: false})`

delambo:
Thanks for reporting this @bkniffler. Can you give us an example of what you're fixing here. A pull request would be welcome too.

bkniffler:
If there is an error on the server side, the stack-traces will refer to the compiled main.js file since nodejs is not able to pickup the source map, hence refer the original code. https://www.npmjs.com/package/source-map-support is fixing this.

tizmagik:
The BannerPlugin would put it at the top of every file. Is that what we need or do we just need it at the top of the server entry file? (`{entryOnly: true}`)

delambo:
@bkniffler - I had to use the new one-argument options method and define `banner` to get this to work:
```
new webpack.BannerPlugin({ banner: 'require("source-map-support").install();', raw: true, entryOnly: false }),
```

delambo:
Also, I'm testing this against errors in our internal app and it's not really adding anything. I'm adding `throw 'error'` in random places around the code and I get stack traces with or without this change. What use cases did you find that this addition fixed?

delambo:
@bkniffler - Nevermind. I forgot that we were using `source-map-support` in our internal app so I was able to get a good use case 🙄

bkniffler:
Ah, was just about to file in a testcase ;) Just in case, a `server/index.js` file like this might be enough:

```
import express from 'express';

const app = express();
throw new Error('Test');
app.listen(parseInt(KYT.SERVER_PORT, 10));
```

PR #402, opened 2017-02-16
add source-map-support to webpack config

bkniffler:
https://github.com/NYTimes/kyt/issues/397

### DIFF
--- a/packages/kyt-core/config/webpack.dev.server.js
+++ b/packages/kyt-core/config/webpack.dev.server.js
@@ -53,6 +53,6 @@ module.exports = options => ({
   plugins: [
     new webpack.NoErrorsPlugin(),
     new webpack.optimize.LimitChunkCountPlugin({ maxChunks: 1 }),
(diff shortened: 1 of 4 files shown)


packages/kyt-core/config/webpack.dev.server.js
```diff
@@ -53,5 +53,6 @@ module.exports = options => ({
   plugins: [
     new webpack.NoErrorsPlugin(),
     new webpack.optimize.LimitChunkCountPlugin({ maxChunks: 1 }),
+    new webpack.BannerPlugin('require("source-map-support").install();', { raw: true, entryOnly: false }),
```
delambo: I think you need to use the one option argument: 

`{ banner: 'require("source-map-support").install();', raw: true, entryOnly: false }`

bkniffler:
Okay, done. I've also set `entryOnly: true` which should be sufficient.

delambo:
Thanks for the help @bkniffler.

zjye:
@bkniffler , in this case, will need to bring over the 'source-map-support' in node_modules together with the bundled js when deploy to server? or else it can't find the module when running with node

Issue #401, opened 2017-02-15
Did 0.4.0 break testing local starter kyts?


janhoogeveen:
### Versions of kyt, node, OS
kyt 0.4.0 after following the migration guide

### Bug or Feature?
Bug

### Current Behavior
in 0.3.0 I was able to use the bootstrap command and test a local starter-kyt, this made sure my Github only shows clean commits. I updated to kyt 0.4.0 and want to test my changes before pushing them to Github. As I understand it the transition to a monorepo with kyt-cli means we don't need to install the monorepo and use bootstrap commands anymore to perform command line magic.

These are the 2 scenario's I've tried.
First, installing it with a globally installed kyt-cli

```
✔ 17:24 ~/Projects/cleverfranke $ kyt-cli setup -d test -r kyt/.git --repository-path ./cf-kyt-starter-universal-redux/
0.19.1

🔥  Let's set up your new kyt project...

✨  Answer a few questions to get started  ✨

? Choose an installer yarn
👍  Creating your new project at test
👍  Setting up the specified starter-kyt
fatal: repository 'kyt/.git' does not exist


❌  There was a problem cloning the repository

fatal: repository 'kyt/.git' does not exist


❌  Failed to setup: kyt/.git
```

Second try, a locally installed kyt-cli

```
✔ 17:35 ~/Projects/cleverfranke/cf-kyt-test [master L|…2261] $ node_modules/.bin/kyt-cli setup -r kyt-cli/.git --repository-path ../cf-kyt-starter-universal-redux
0.19.1

🔥  Let's set up your new kyt project...

✨  Answer a few questions to get started  ✨

? Choose an installer yarn
? Enter a new directory name. To install in current directory, leave blank.
👍  Setting up the specified starter-kyt
fatal: repository 'kyt-cli/.git' does not exist


❌  There was a problem cloning the repository

fatal: repository 'kyt-cli/.git' does not exist


❌  Failed to setup: kyt-cli/.git
```

### Desired Behavior
After installing kyt-cli (globally or locally) I want to be able to pinpoint a local directory and use that as a starter kyt.

janhoogeveen:
Worth mentioning is the fact that I would be fine with specifying a specific branch/commit/tag as well when using the kyt-cli setup command and just setup from a public repo. For example, this would be a-okay in my book:

```
kyt-cli setup -r https://github.com/cleverfranke/cf-kyt-starter-universal-redux.git#feature/migrate-0.4.0
```

[...]

PR #403, opened 2017-02-16
Adds local-path argument for testing local starter-kyt setup

ccpricenytimes:
This is to address #401

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -73,9 +73,11 @@ It is a common workflow to make changes to kyt-core and test them with `kyt-cli 
 
 ### Testing local starter-kyt changes
 
-To test setting up/installing local starter kyts, you need to specify the `--repository-path` option. The following command will create a test directory and install the given local kyt repository and reference the starter kyt using the given repository path.
+To test setting up/installing local starter-kyts, you need to specify the `--local-path` option. This allows you to install a starter-kyt locally by copying rather than cloning from git.
 
-`kyt-cli setup -d test -r kyt/.git --repository-path packages/kyt-starter-static`
+`--local-path` should be a path written relative to where you run your command
+
+`kyt-cli setup -d test  --local-path /path/to/my-local-starter-kyt-dir`
 
 ### Testing kyt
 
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -73,9 +73,11 @@ It is a common workflow to make changes to kyt-core and test them with `kyt-cli
 
 ### Testing local starter-kyt changes
 
-To test setting up/installing local starter kyts, you need to specify the `--repository-path` option. The following command will create a test directory and install the given local kyt repository and reference the starter kyt using the given repository path.
+To test setting up/installing local starter-kyts, you need to specify the `--local-path` option. This allows you to install a starter-kyt locally by copying rather than cloning from git.
 
-`kyt-cli setup -d test -r kyt/.git --repository-path packages/kyt-starter-static`
+`--local-path` should be a path written relative to where you run your command
+
+`kyt-cli setup -d test  --local-path /path/to/my-local-starter-kyt-dir`
```
jaredmcdonald: but an absolute path (like the one above) will work too, right? if so, i would drop the "written relative to where you run your command", that seems like it can be assumed

packages/kyt-cli/cli/actions/setup.js
```diff
@@ -539,12 +546,16 @@ module.exports = (flags, args) => {
...
-    if (!args.repository) questions.push(skQ);
+    if (!args.repository && !localPath) questions.push(skQ);
 
     inquire.prompt(questions).then((answer) => {
       // question 1
       ypm = answer.ypm || ypm;
       // question 2
+      // Save Local directory Path before moving to new directory
+      if (localPath) {
+        localPath = path.join(process.cwd(), localPath);
```
jaredmcdonald: actually this answers my question above, this seems counterintuitive. why not just allow any path? if you join it to the cwd, then you can only use a relative path

```
> path.join(process.cwd(), '/usr/bin/')
'/<my_cwd>/usr/bin/'
```

how about using `path.resolve` instead?

```
> path.resolve('/usr/bin')
'/usr/bin'
> path.resolve('../usr/bin')
'/Users/usr/bin'
```

jaredmcdonald:
another thought - this seems like it could be useful outside of just testing, so maybe we can keep it (mostly) undocumented for now with the intention of making it a full feature later?

CONTRIBUTING.md
ccpricenytimes: yep. done!

PR #405, opened 2017-02-20
Adds more functional tests

delambo:
- adds functional tests for `kyt test`
- adds functional tests for `kyt build`
- adds functional tests for universal and static starter kyts. note, I used snapshots to test against the html output of the dev/build server. The one drawback is that we might have to update the snapshots often because the assets hashes will change after every dependency update. Maybe it will be a good thing.
- changes info logging when using default babelrc. It was logging three times after running `kyt test`, once for the client config, server config and jest config. I moved the logging into `buildConfigs` so it only happens once for `test` and the other commands.


e2e_tests/fixtures/util.js
```diff
@@ -0,0 +1,24 @@
...
+const path = require('path');
+const rootDir = process.cwd();
+
+shell.config.silent = true;
+
+module.exports = {
+  setupStageWithFixture: (stageName, fixtureName) => {
+    const stagePath = path.join(rootDir, stageName);
+    shell.mkdir(stagePath);
+    shell.exec(`cp -a ${rootDir}/e2e_tests/fixtures/${fixtureName}/. ${stagePath}/`);
```
jaredmcdonald: `path.join`, if we think anyone will be developing kyt on windows (?)

delambo: I tried super hard to get it to work with `shell.cp` but it wouldn't actually copy the files `/.` into the directory. :man_shrugging:

PR #408, opened 2017-02-21
Pass through webpack instance to modifyWebpack functions

jaredpalmer:
This passes kyt's version of webpack to the user's `modifyWebpack` function as another argument. The user thus does not need to install webpack themselves for simple things like adding a postcss plugin or another webpack flag.

### DIFF
--- a/docs/Recipes.md
+++ b/docs/Recipes.md
@@ -39,7 +39,7 @@ if (options.type === 'client') {
 In `kyt.config.js`
 
 ```javascript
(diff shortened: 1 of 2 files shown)


delambo:
Hey @jaredpalmer - For now, I think we're going to advise users to do something similar to #410. Since `webpack` is a dependency of `kyt`, it is hoisted to the top level of the `node_modules` directory and is accessible via `require`. The goal is for kyt to become more opinionated over time so that users will never have to override the config. Thanks for your time.

Issue #14, opened 2016-08-10
Babel support for kyt.config.js


tizmagik:
Right now, kyt.config.js can only be written in node v6 supported ES6 (e.g. can't `import` modules). It would be nice to allow full ES6+ support in the config.

julianvmodesto:
Would this be as simple as [`require('babel-register')`](https://babeljs.io/docs/usage/babel-register/) in kyt-core/utils/kytConfig.js?

PR #409, opened 2017-02-22
Use babel-register in kytConfig

julianvmodesto:
### Issue

Attempt at #14 

### Solution

Added babel-register to kytConfig which will babelify the `require` for the user's kyt.config.js.

Quickly tested with an `import path from 'path'` in a test kyt project and it seemed to work out.

Maybe we could use the babel-register `only` directive to target only the kyt.config.js? Or we can move babel-register up and babelify more of kyt-core.


julianvmodesto:
Oops, broken build. `BABEL_DISABLE_CACHE=1 yarn run test` runs successfully, but going to investigate what's failing here.

julianvmodesto:
Closing for now but will come back to this...

PR #410, opened 2017-02-22
Added missed require statements

Skorney:


delambo:
Thanks for catching that @Skorney! I'm sure your change will help other users.

PR #418, opened 2017-02-24
fix: changed dirname and filename config in webpack configs

maxrbaldwin:
`dirname` and `filename` are both currently set to false in both `prd` and `dev` server configs. Although in `webpack.base.js` they are set to true. When these configs are set to false in the webpack config it changes the relativity of the Node global objects `__dirname` and `__filename`. When they are set to false  `__dirname` and `__filename` will return paths relative to the `build` directory. When these configs are set to true the will return paths relative to the file that you call them in.

**Screenshots:**
- console log `__dirname` in `src/server/index.js` with configs set to false: https://cl.ly/0D3k1C2K0g3k
- console log `__dirname` in `src/server/index.js` with configs set to true: https://cl.ly/1F371r0X1k20

### DIFF
--- a/packages/kyt-core/config/webpack.dev.server.js
+++ b/packages/kyt-core/config/webpack.dev.server.js
@@ -18,8 +18,8 @@ module.exports = options => ({
   target: 'node',
 
   node: {
-    __dirname: false,
-    __filename: false,
+    __dirname: true,
+    __filename: true,
   },
 
   externals: nodeExternals(),
--- a/packages/kyt-core/config/webpack.prod.server.js
+++ b/packages/kyt-core/config/webpack.prod.server.js
@@ -22,8 +22,8 @@ module.exports = options => ({
   target: 'node',
 
   node: {
-    __dirname: false,
-    __filename: false,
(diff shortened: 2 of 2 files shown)


jaredmcdonald:
hey @maxrbaldwin, thanks for the PR! on first glance this seems like an intentional choice to me--i can't see why baking the source structure into the server bundle would be desirable; rather, it seems intuitive to have `__dirname` resolve at runtime on the server (@delambo or @ccpricenytimes, let me know if i'm wrong there). can you expand a little bit on your use case and maybe we can find another way to deal with it?

maxrbaldwin:
Hey @jaredmcdonald, at first I thought the same thing as well. It seems strange to want to change these global values. 

My use case is pretty standard. I was simply constructing a full file path using `path.join`. `__dirname` is being used as my file root. I kept getting an error because of the incorrect file root. Although the answer to this issue was easy to find, I can see how this can be confusing to users of KYT.

One thing to note, `__dirname` resolves as you would expect it to in `kyt.config`. This issue is for any reference to `__dirname` in a project.

Issue #399, opened 2017-02-14
RFC: Versioning supported starter-kyts with npm


ccpricenytimes:
## Summary

We should version and publish supported starter-kyts to npm so they can be kept in sync with their version of kyt and changes can be made to the codebase without breaking starter-kyts.

## Motivation

Since we've moved to a monorepo it would be great to make changes to both kyt and starter-kyts simultaneously (eg. when updating to a new version of webpack) but we currently cannot do this because starter-kyts are installed from the master branch of the code base and kyt is versioned through npm. 

## Detailed design

Supported starter-kyts would be versioned using semver and published to npm. Each starter-kyt would continue to use the "kyt" section of their `package.json` to specify which version(s) of kyt it supports. kyt-cli would download these starter-kyts from npm rather than through git clone. kyt-cli would always download the lastest version of the starter-kyt. 

Rather than cloning the github repo, the `setup` command will npm install the starter-kyt into the temporary directory. From there steup will continue in the same way it currently does. 
Users will still be able to install starter-kyts from other repos using the `-r` flag. This will continue to use git clone.

This will allow us to make changes to the starter-kyt code without breaking any existing setups. 

We will package the starter-kyts in their own package.json so npm install won't install the numerous dependencies but rather can just be used to clone. 

## Drawbacks
This requires versioning the starter-kyts and managing more npm packages. I think the benefits of being able to make changes to the entire monorepo outweighs this. 

## Alternatives
1. We could continue to use github and download the starter-kyts off of some sort of tag system. The problem here is that we would have to figure out how to always pull the latest tag through kyt-cli so we wouldn't have to publish a new version kyt-cli every time a starter-kyt changes.  
3. we can continue with the status quo which requires multiple branchs for changes that affect both kyt and the starter-kyts.

janhoogeveen:
I'm all in favour of versioning both kyt and starter-kyts! I'm working on my own starter-kyt and publishing a package to NPM is something I've never done before but I'm sure I'll manage.

Installing the latest version of a starter-kyt makes sense as a default, although I feel like eventually it would be useful if you can specify which starter-kyt version to install with kyt-cli. Though I suspect that's a very low priority enhancement which I'm not even sure I would use myself.

ccpricenytimes:
[...]

PR #425, opened 2017-03-02
Version and install starter-kyts with npm

ccpricenytimes:
connects #399


packages/kyt-cli/cli/actions/setup.js
```diff
@@ -414,10 +414,25 @@ module.exports = (flags, args) => {
     removeTmpRepo();
```
jaredmcdonald: minor but can we rename variables to not refer about "repos"? kind of confusing now since we won't use repos exclusively to manage starter-kyts after this

packages/kyt-cli/cli/actions/setup.js
```diff
@@ -414,10 +414,25 @@ module.exports = (flags, args) => {
...
-      shell.cp('-r', localPath, tmpRepo);
-      afterClone();
+      shell.exec(`cp -R ${localPath} ${tmpRepo}`);
+      afterCopy();
+    } else if (npmName) {
+      shell.mkdir(tmpRepo);
+      shell.cd(tmpRepo);
+      const fakePkgPath = `${tmpRepo}/package.json`;
+      fs.writeFileSync(fakePkgPath, JSON.stringify(fakePackageJson, null, 2));
+      let iCmd = 'npm install';
```
jaredmcdonald: does this let you install a previous version of a starter-kyt, or just the latest on npm?

packages/kyt-cli/cli/actions/setup.js
ccpricenytimes: yeah definitely!

packages/kyt-cli/cli/actions/setup.js
ccpricenytimes: just latest for now. I think we want to have people install latest, but we could talk about supplying versions in the future if we see a use case

jaredmcdonald: hmm i think it would probably just work if you supplied an `npmName` like `foo@0.1.0`, right?

related: can we make it so that you can supply an `npmName` on the command line?

PR #436, opened 2017-03-11
Adds Clean script

tizmagik:


Skorney:
This `clean` and `bootstrap` scripts looks similar to `lerna clean --yes|bootstrap` and `lerna bootstrap` commands. Is it any reason to write own `bootstrap` and `clean` scripts? 

I'm using kyt with lerna.js in our projects, it seems like it works pretty good.

delambo:
I was thinking about spiking on Lerna. We probably won't use 90% of Lerna's functionality since we're an `--independent` project and I don't really care for all of the lerna->npm run proxying and publishing stuff. If anyone wants to draw-up a quick POC I'd love to see what it looks like.

I'm going to merge this in the meantime.

PR #440, opened 2017-03-14
version bumps for babel presets

jaredmcdonald:


packages/kyt-core/package.json
```diff
@@ -22,7 +22,7 @@
     "babel-jest": "19.0.0",
     "babel-loader": "6.3.2",
     "babel-polyfill": "6.16.0",
-    "babel-preset-kyt-core": "0.1.0",
+    "babel-preset-kyt-core": "0.2.0-rc.1",
```
jaredmcdonald: i guess this probably means `kyt-core` needs a version bump too, right? @delambo

ccpricenytimes: yep.

Issue #349, opened 2016-12-15
Webpack 2.2 RC (Performance Budgets)


tizmagik:
## Webpack 2.2 RC

Webpack 2.2 RC includes support for Performance Budgets (also, dynamic imports, etc). However, kyt currently silences webpack's output so users won't be able to take advantage of this feature.

Read more here: https://medium.com/webpack/webpack-performance-budgets-13d4880fbf6d#.da4lr9175

### Bug or Feature?

Feature

### Current Behavior

Webpack output is swallowed

### Desired Behavior

Webpack output is (optionally?) visible

### Potential Solutions

1. Include an option in `kyt.config.js` to bubble up/opt for Webpack's output
2. Remove kyt's current output handling and leave it up to Webpack

I'd suggest `#2` because Webpack's output is much richer than what we currently show. We can still emojify the start/end to communicate clearly to the user what's going on task-wise, but otherwise muting webpack's output seems a disservice, especially given the new Performance Budget feature.

Thoughts?

nahue:
+1 on optionally show webpack's output

nahue:
any news on this?

tizmagik:
I think the plan is to release kyt 0.4.0 and then look in to upgrading Webpack since there are some potentially breaking changes there.

delambo:
We could probably reach into the webpack stats and surface the performance budget to the user so that we can avoid all of the other noise that webpack logs. But I'm not sure if kyt is going to need this feature soon. After we land vendor bundling and work on a user manual that better explains the goals behind code splitting in React apps, and since most users probably create projects from our starter kyts, most kyt projects won't need the warnings that this tool provides. Osmani's [original rationale behind this feature](https://medium.com/webpack/webpack-performance-budgets-13d4880fbf6d#.5t3knpuyp) was to warn users that were likely building a single dist: "we noticed quite a few folks were shipping down monolithic JavaScript bundles." I don't think this will apply to most kyt projects, especially if we're doing a good job of guiding our users.

tizmagik:
Just to follow up here: as discussed in person, it still makes sense to have the option of showing the webpack output (for things like this performance budget and otherwise) because it is still possible to ship down a large JS chunk even if you're code-splitting at the route level. In those cases, having the performance budget warning would indicate to users to code-split _within_ a given route, etc.

PR #448, opened 2017-03-25
fixes warning logging

delambo:
In our internal project, we dump all of the stats when there is a warning in the webpack build which is overly verbose and confusing. This fixes that by just logging the actual warnings.


tizmagik:
Haha, FWIW I was using this is a workaround until #349 landed 😁

Issue #444, opened 2017-03-21
Allow CSS modules selectors


janhoogeveen:
### Versions of kyt, node, OS
0.4.x

### Bug or Feature?
Bug

### Current Behavior
When using CSS modules stylelint will throw an error when using :global or :local selectors.

### Desired Behavior
:global and :local selectors should be an exception in the selector-pseudo-class-no-unknown rule.

```js
{
  "extends": "stylelint-config-kyt",

  "rules": {
    "selector-pseudo-class-no-unknown": [
      true,
      {
        "ignorePseudoClasses": ["global", "local"],
      }
    ]
  }
}
```

PR #454, opened 2017-03-28
Updates stylelint-config-kyt allowing use of CSS Modules pseudo selectors

janhoogeveen:
This should fix issue #444 

Note: The lockfiles aren't updated yet. I'm guessing the update to stylelint-config-kyt needs to be published before yarn can find them?


jaredmcdonald:
@delambo i think it's trying to install 0.3.1 on travis, should we pre-publish it from the branch or something? or roll back the version change and bump the version once this hits master?

delambo:
@janhoogeveen 🙏 🙏 🙏

Issue #457, opened 2017-03-30
Add support for connect-history-api-fallback?


douglasback:
### Versions of kyt, node, OS

Latest!

### Bug or Feature?
Feature! (Request? But also I can have a PR for this as I've done half of the work on a fork.)

### Current Behavior

I started a static client-side only project created based on the kyt-starter-static starter kyt. 

I'm relying on React Router for any and all navigation. Currently, when working locally (`npm run dev`) if I navigate to a page other than the root (i.e. from http://localhost:3001/ to http://localhost:3001/something), reloading the page gets a 404 as the express server can't find assets to serve from `/something`.

### Desired Behavior

Ideally, refreshing the page on a route doesn't 404! If the Connect server started by kyt's [dev action](https://github.com/bripkens/connect-history-api-fallback) took advantage of [a history api fallback](https://github.com/bripkens/connect-history-api-fallback) middleware, this wouldn't happen.

This could take a couple of different forms, but the simplest one is to extend add a second line to [this](https://github.com/NYTimes/kyt/blob/master/packages/kyt-core/cli/actions/dev.js#L47) making it look like:

```
if (!hasServer) {
  app.use(history(historyMiddlewareOptions));
  app.use(express.static(publicSrcPath));
}
```

The middleware can also accept a variety of options; allowing those to be defined in the kyt.config.js would make them accessible here.

As I mentioned, I've done half of the work here in a fork and would be happy to contribute a PR. But also if this is possible today without making changes, then let me know!

delambo:
Hey, Douglas - it's great to see you around here! Sorry you ran into this problem; it's not documented well. I figured folks would setup nginx or build a simple one-page/route-less app.

That said, I think this might be a good change for `dev`. What do you plan on doing with your production `build`? Would you run nginx in front of your app and route everything to the root html file?

douglasback:
Ah yes, important context here — the use case for production is that the app is served from an S3 bucket, which is fronted by Varnish (any CDN would probably do here). Varnish's configuration would serve the root HTML file from any path, allowing the Router to determine which components to render.

delambo:
Ah, cool. I think this would be a good addition to `dev`. Push up on your fork when you get a chance. As a followup we should add some docs about this magic and a note about how to prepare for production.

delambo:
this made it out in 0.5.3

PR #458, opened 2017-04-06
Add connect-history-api-fallback middleware for client-only apps in dev

douglasback:
Per the chat in #457 — opening this as a conversation starter. 

I didn't add docs yet, in part because I'm not sure where the best place for them is.

The use case for this is that I have a client-side SPA that in staging and production has a reverse proxy that resolves all URL requests to the entry page (published at index.html). React Router takes care of the routing from there. The history-api middleware here accomplishes something very similar.


delambo:
🙏

PR #462, opened 2017-04-10
Removing at-no-rule-unknown

ccpricenytimes:
In upgrading our internal project I noticed that the `at-no-rule-unknown` rule was showing errors for standard sass syntax eg. `@include`, `@mixin` etc. 
It seems to brittle to support all sass syntax like this as exceptions so this is a PR to remove the rule all together. 

The rule can be added to user's local stylelint configs if they wish to keep it. 

cc @janhoogeveen would love your feedback on this change. Thanks!

### DIFF
--- a/packages/babel-preset-kyt-core/yarn.lock
+++ b/packages/babel-preset-kyt-core/yarn.lock
@@ -18,99 +18,99 @@ babel-code-frame@^6.22.0:
     esutils "^2.0.2"
     js-tokens "^3.0.0"
 
-babel-helper-builder-binary-assignment-operator-visitor@^6.22.0:
-  version "6.22.0"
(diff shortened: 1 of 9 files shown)


janhoogeveen:
@ccpricenytimes Yikes! Didn't anticipate this issue!

My best guess is hardcoding the rules in the config like I did overwrites the default settings. I'm probably gonna need to find a way to extend these settings. 

I'd like to try if this plugin for instance solves the problem in a cleaner way: https://github.com/pascalduez/stylelint-config-css-modules

That said I strongly believe it would be best if kyt-core solves this instead of placing it in the user config, since css-modules are supported by Kyt.

janhoogeveen:
Ah I misread the issue and checked the diff. If removing that specific line fixes your issues that looks fine to me. The biggest win for css-modules support is the line above it allowing the usage of global scoping as an opt-in.

PR #473, opened 2017-04-26
move react-hot-loader after babel-polyfill

dkaoster:
Because of an ie11 bug, 'react-hot-loader/patch' needs to come after 'babel-polyfill'.

https://github.com/facebook/react/issues/8379#issuecomment-273489824


ccpricenytimes:
Hey @dkaoster this looks great. Were you able to test locally? I don't have an IE11 VM

dkaoster:
@ccpricenytimes yup, works locally on my machine.

PR #480, opened 2017-05-03
Add Prettier + ESLint Integration

### LINKED ISSUE #445: Use prettier instead of linting
### Versions of kyt, node, OS
Any future version

### Bug or Feature?
Feature

### Current Behavior
We lint with ESLint. It's nice enough but tedious.

### Desired Behavior
There's a lot of buzz around Prettier and how it helps people focus on writing code instead of adhering to style rules. I'd be interested in auto-formatting code with Prettier as part of the stack.

kohlmannj:
fixes #445

This PR adds the following enhancements:

- kyt-core: add [Prettier](https://github.com/prettier/prettier), [eslint-config-prettier](https://github.com/prettier/eslint-config-prettier), and [eslint-plugin-prettier](https://github.com/not-an-aardvark/eslint-plugin-prettier) as dependencies
- eslint-config-kyt: add the above three modules as peer dependencies
- eslint-config-kyt: update `.eslintrc.json` to configure Prettier for 100-character line length, single quotes, and ES5-style trailing commas (which I think is the closest approximation of the existing `comma-dangle` rules)
- eslint-config-kyt: update Install section of the README to note the additional dependencies

Note: `eslint-config-prettier` disables ESLint rules that might conflict with Prettier, while `eslint-plugin-prettier` runs Prettier (and passes it a config) as part of the linting process.

Also, ESLint and its ruleset covers more than just code formatting, so it's appropriate to use both ESLint and Prettier _together_, rather than replace one with the other.


kohlmannj:
I'm not sure what additional dependencies I need to add such that the Travis build will work again. I thought `kyt-core` was the only place where I needed to add these new dependencies…

packages/eslint-config-kyt/eslintrc.json
```diff
@@ -42,6 +44,11 @@
       "imports": "always-multiline",
       "exports": "always-multiline",
       "functions": "never"
+    }],
+    "prettier/prettier": ["error", {
+      "printWidth": 100,
```
staylor: `"semi": true,`

delambo:
🙏  @kohlmannj

PR #486, opened 2017-05-23
update eslint-plugin-jsx-a11y to 5.0.3

dkaoster:
Updates the eslint-plugin-jsx-a11y to include updated rules. We'd like to use this one specifically:

https://github.com/evcohen/eslint-plugin-jsx-a11y/issues/91


delambo:
@dkaoster - sorry for getting to this so late. The plugins were updated recently. Please come back and help with other stuff in the future 🙏

PR #527, opened 2017-08-11
Update node-sass to latest version

janhoogeveen:
The current version of node-sass has no bindings available for Node 8. Since Node 8 is going to be LTS pretty soon it might be nice to upgrade to node-sass. It even has the added benefit of being able to run the universal app on an Alpine environment.


delambo:
Thanks again @janhoogeveen 🙏

janhoogeveen:
No problem at all! Hope the lockfile isn't problem, since I see I didn't update that one.

delambo:
np, I'll generate the lock and add it to the 0.9.0-alpha.4 as soon as we merge alpha.3 in.

Issue #511, opened 2017-07-11
Remove yarn support


delambo:
Modern versions of npm support lock files and have comparable speeds to yarn. Also, yarn has a different api and implementation as we saw in #510. We should move kyt off of yarn and only support npm.

### Current Behavior

Currently, kyt-cli supports yarn and npm, and the kyt development process relies on yarn.

### Desired Behavior

Remove yarn support and recommend npm 5.0+.

Jasu:
Hi,

We use yarn for it's support for [workspaces](https://yarnpkg.com/blog/2017/08/02/introducing-workspaces/) and [resolutions](https://yarnpkg.com/lang/en/docs/selective-version-resolutions/). I don't think NPM has either.

delambo:
Thanks for the info @Jasu. I think we can close this now. I removed yarn from our bootstrap process, and I think it's ok to leave support for users in the CLI.

Issue #437, opened 2017-03-11
ESLint in editor (Atom) not working


tizmagik:
### Bug or Feature?
Bug

### Current Behavior
![image](https://cloud.githubusercontent.com/assets/244704/23824293/d119cab0-0641-11e7-9cf0-c45d84c8805f.png)

### Desired Behavior
ESLint in editor works.

Noticed this in Atom, have not seen/tried other editors.

delambo:
I think this is because we're using the eslint dependencies in kyt-core. My original idea was to dog-food kyt-core's linter and keep the dependencies in one place but if we want our text editors to lint we'll have to additionally manage the eslint dependencies in the root package.json [sigh]. 

Unfortunately, lerna won't help us with this since it only hoists common devDependencies to the top level but let me know if I'm wrong because I've been thinking about a spike on lerna.

tizmagik:
Yea, lerna wouldn't help here, but I think managing the eslint dependencies in the root of package.json wouldn't be so bad right? Isn't it just a matter of adding eslint, eslint-config-airbnb et al to devDependencies?

delambo:
Yeah, it's not terrible, I'm just being overly idealistic. Since we're in the game of dependency management and we're pretty aggressive about updates I thought it would be simpler to maintain things in one place. NBD.

delambo:
This works now since we use lerna to hoist everything to the root on bootstrap.

PR #530, opened 2017-08-16
Removes yarn dev support; Adds lerna

delambo:
This removes the yarn dev support outlined in #511. In place of our bootstrap script, I added [Lerna](https://github.com/lerna/lerna). 

With lerna's newish `--hoist` flag, I hoist all the monorepo dependencies to the root `node_modules` directory. Because of dependency problems in the starter kyts, I had to manually install dependencies in the `starter-src` directories and link kyt-core. Since most changes in kyt are dependency upgrades, I added a `clean-bootstrap` which will hopefully clean/hoist dependencies after an upgrade/pull.

I haven't been able to test publishing but I did add the following commands which I hope will be helpful: `publish-next`, `publish-release`. It might get tricky since we have the nested `starter-src` package directories with local dependencies.

I'm pretty sure we'll have to tweak the lerna configs as we do real work in the repository. TBD.


tizmagik:
Great! Do you know if this helps with https://github.com/NYTimes/kyt/issues/437 or the babel-preset-env PR that's been suck?

delambo:
@tizmagik - I'm not sure yet, but hopefully. I'll test that soon.

PR #531, opened 2017-08-28
Adds babel-preset-env

delambo:
Replaces `babel-preset-latest` with `babel-preset-env` in `babel-preset-kyt-core`.

Because the lerna hoisting doesn't work for all dependencies, I added some hoisting exceptions to the lerna.json configuration and a script that installs the starter kyt dependencies but keeps the lerna symlinks. That way, starter kyt commands like `kyt dev` can work against the local repository. One drawback to the hoisting that I haven't figured out is that `kyt test` does not work against the starter kyts, but I'm not terribly concerned about that since the starter kyt unit tests are for show.


tizmagik:
As a follow-on PR (or maybe as part of this one) we should also set `BABEL_ENV` to `server` or `client` for the 2 builds. Server can be set to `"node": "current"` and the client to `"browsers": ["last 2 versions", "safari >= 7"]` (or whatever as a good default).

PR #553, opened 2018-04-05
WIP: Update to webpack 4.5

ctborg:


erchaves:
Will this upgrade improve performance of the build process? I have noticed that the build process is kind of slow for my project. I haven't yet dug into how it might be improved, but I might be interested in helping with that. I'm curious if that is a motivation here for switching to mini-css-extract-plugin...

PR #554, opened 2018-04-16
prevent kyt to hang if webpack config run fails

thiagopnts:
the try-catch  gets only config validation errors, if something goes wrong after that, e.g. EACCESS, etc kyt will hang waiting for the `done` hook that will never be called


packages/kyt-core/utils/webpackCompiler.js
```diff
@@ -7,7 +7,12 @@ module.exports = (webpackConfig, cb) => {
 
   // Compile the webpack config
   try {
-    webpackCompiler = webpack(webpackConfig);
+    webpackCompiler = webpack(webpackConfig, err => {
+      if (err) {
+        logger.error(`${type} webpack error:\n`, err);
+        process.exit();
```
tizmagik: Should this be an exit code of 1?

delambo: Yes good call!

delambo:
As you can see in the starter kyt end to end test, something weird is going on when starting up the dev server. I'll try to debug some more later.

Note on how to reproduce:
```
npm run clean-bootstrap
cd packages/kyt-starter-universal/starter-src
node_modules/.bin/kyt dev
# Should yield:  ❌ Server build failed
```

delambo:
I couldn't get this to work, per above. I'm going to close for now. If anyone wants to give it a try, please reopen.

PR #558, opened 2018-05-22
Adds babel cache in development

delambo:
We removed the babel cache directory after we encountered problems in production. It turns out, now that we have a large codebase, we really need it in development. It cuts our dev startup time down by ~35% (after the first time), and speeds up incremental builds by ~10%.

By default, the cache directory is stored in node_modules, but since that gets blown away occasionally developers would need to prime their caches often. To get around that problem, I opted to use a temp directory on the OS.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,5 @@ test-packages
 
 # Local stuff
 packages/kyt-starter-*/starter-src/build
+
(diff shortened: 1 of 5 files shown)


callahanchris:
Nice change, thanks @delambo 👍

Issue #559, opened 2018-05-23
Update react-router to the latest version (React Router 4)


erchaves:
# RFC: Update react-router to the latest version (React Router 4)

## Summary
[summary]: #summary

React Router is an important part of the react ecosystem. KYT currently uses a significantly older version of react router. In order to keep KYT up to date and compatible with future development, this RFC proposes that we update to react router 4.

## Motivation
[motivation]: #motivation

Rect router 4 is significantly different from older versions. I think react 4 is better and simpler to understand. As demonstrated in the corresponding PR, I believe it also fixes issues that otherwise had to work around, like this [react-hot-loader issue](https://github.com/NYTimes/kyt/blob/9fd7212d2e9edde6c37ac083e1c5491a113bb2de/packages/kyt-starter-universal/starter-src/src/routes/index.js#L38).

## Detailed design
[design]: #detailed-design

This PR upgrades react-router 2.6.1 with react-router-dom 4.2.2. React-router-dom is just the flavor of react-router with DOM bindings, and corresponds to react-router 4.2.X.

The PR only focuses on the kyt-starter-universal package for now, but these changes could be propagated to other packages as needed.

Other files changed include:
`client/Root.js`
`client/index.js`
`components/App/index.js`
`routes/index.js`
`server/index.js`

The server side code in `server/index.js` is updated to follow the [recommended approach](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/StaticRouter.md) for matching routes in react router 4. 

## Drawbacks
[drawbacks]: #drawbacks

While I have already done the work of making these updates and running the tests, this will take a little work from the team to vet this update and make sure it's compatible with all the packages.
## Alternatives
[alternatives]: #alternatives

We could leave the react-router version as is. The downside is that KYT could start to feel out-of-date as future developers will be accustomed to the newer react router format. They could be potentially confused by the old react router api.

## Unresolved questions
[unresolved]: #unresolved-questions

- The update to react-router 4 in the universal package probably needs to be propagated to the other packages.
- This PR removed some code-splitting related code in routes/index.js which seemed unnecessary. I'm not sure if that is still necessary and desired and should be added back in.

erchaves:
As mentioned in the unresolved questions, I'm not sure if this change is compatible with code-splitting. This article is a good reference for how we could maybe implement automatic code-splitting with react router 4. https://serverless-stack.com/chapters/code-splitting-in-create-react-app.html

[...]

PR #560, opened 2018-05-23
React router upgrade

erchaves:
## Update react-router to the latest version (React Router 4)

React Router is an important part of the react ecosystem. KYT currently uses a significantly older version of react router. In order to keep KYT up to date and compatible with future development, this RFC proposes that we update to react router 4.

For more details, please see [RFC 559](https://github.com/NYTimes/kyt/issues/559)


erchaves:
Hmm, it looks like the e2e test is failing at:
`  ● starter kyts › kyt-starter-universal › should start a dev server on :3000`

I'm not sure how to debug that at the moment. When I use `kyt-cli setup --local-path` to make a copy of the updated package, it builds and runs fine for me, and the tests pass.

delambo:
Hey, Eric - I think the easiest way to test something like this is to run the e2e tests locally from the root directory: `npm run e2e`. To make it easier, you might want to comment out the other tests in `e2e_tests/tests/*.js` or specify the test you want to run in: `e2e_tests/jest.config.json`.

We're slowly working toward RR4, as well. We haven't had a lot of time or need for it, so it will be interesting to see how this works.

erchaves:
Hi @delambo, thanks for the response. I figured out the issue with the tests and got it fixed.  

As referenced above, here is the accompanying [RFC 559](https://github.com/NYTimes/kyt/issues/559). Here's a short summary: 
- This updates us to the latest react router, keeping us up to date with their documentation.
- It removes the need for the react-hot-loader work around.
- A possible down side is that it removes the code-splitting code. I'm not sure if that is a priority to keep included in the default package. If it is, the RFC discusses how that could possibly be remedied through a little more investigation. 

If this feels like the right approach, let me know if I can assist further. I'm hoping to use kyt on a number of projects including some open source work for museums. I can share more details about that as well if you're interested.  Thanks!

packages/kyt-starter-universal/starter-src/src/routes/index.js
```diff
@@ -1,43 +1,58 @@
...
+  children: PropTypes.node.isRequired,
+};
+
+RedirectWithStatus.propTypes = {
+  from: PropTypes.string.isRequired,
+  to: PropTypes.string.isRequired,
+  status: PropTypes.number.isRequired,
+};
+
+export default Routes();
```
staylor: does this need to be a singleton?

erchaves: Good catch @staylor, thx. I've updated this.

delambo:
Again, sorry for the late treatment. @staylor added RR4 in some of his recent work. A 1.0 release will follow, hopefully soon.

PR #562, opened 2018-06-22
updating fs.writeFile to be node >= 9 compatible

saranrapjs:
I was running into this error on node `v10.1.0`. Without an explicit callback, this will throw a `ERR_INVALID_CALLBACK` error.


delambo:
NIce, thanks!

PR #576, opened 2018-09-13
Update node-sass to 4.9.3 from 4.5.3

montmanu:


delambo:
`npm run clean-bootstrap` to update package-lock

montmanu:
should the `kyt-core/package-lock.json` also have updated here, or just the root `package-lock.json`?

tizmagik:
If you're doing it via lerna, it will not update the sub-packages package-lock.

PR #577, opened 2018-10-30
Update Babel deps

staylor:
### What is Babel?
Babel is used to transpile our JavaScript code (ES6) to ES5 + polyfills. This allows us to write modern code while shipping compiled code that the browsers we support understand. This also allows us to use proposed and experimental syntax that Node does not explicity support yet.

## How is Babel configured?
Most projects use a .babelrc file to describe their configuration. This configuration is usually broken up into 3 sections: presets, plugins, and env.

Presets are usually configuration files provided by other npm packages. Plugins are specific pieces of configuration, also provided by other npm packages. env configuration specifies presets and packages that should be used, but only for specfic enviroments, like development, production, and test. Most plugins and presets also allow configuration options to be passed using JSON syntax.
```
{
  "presets": ["@babel/preset-env", "@babel/preset-react"],
  "plugins": ["babel-plugin-add-react-displayname"],
  "env": {
    "test": {
       "plugins": ["babel-plugin-some-package"]
    }
  }
}
```

## When is Babel used?
For VI, our .babelrc is used when Webpack is compiling our source code via babel-loader. VI is an application, so it needs a bundler like Webpack to code-split our app every time it encounters an import() statement. It automatically encounters them when it is traversing various entrypoints in the app, like client/index.js and server/index.js.

For shared-ui packages, like @nyt/components, Babel is invoked via a build CLI command:
```
babel src --ignore "__tests__" -d lib
shared-ui packages are not an app, they are meant to be consumed by other apps
```

## What is KYT?
KYT is collection of tools used to develop, lint, test, and build JavaScript apps. These apps do not have to run on Node JS. KYT supports static site generation a la Next and Gatsby, but it also supports running universal/isomorphic apps on Node. One of the hardest parts about building a JavaScript app today is configuring it. KYT hides as much of this complexity as possible, while giving you flexibility and modern foundation to build your apps upon.

## How does KYT use Babel?
KYT provides 2 Babel presets: babel-preset-kyt-core and babel-preset-kyt-react. For any React app, the most simple Babel configuration will look something like:
```
{
  "presets": ["@babel/preset-env", "@babel/preset-react"] 
}
```
If you app does not use React, then:
```
{
  "preset": ["@babel/preset-env"] 
}
```
Think about KYT's preset as wrappers around each of these presets, but with one advantage. For a KYT React app, you just need to specifiy one preset:
```
{
  "presets": ["babel-preset-kyt-react"] 
}
```
Both presets accept options, and since the KYT React preset includes KYT Core, you can pass options to KYT Core via KYT React:
```
{
  "presets": [
    ["babel-preset-kyt-react", {
      "includeRuntime": true,
      "envOptions": {
        "client": {
          "debug": true
        }
      }
    }] 
  ] 
}
```

## Why do we need to upgrade?
The JS ecosystem is moving towards Babel 7. To take advantage of bug fixes, performance improvements, and new APIs, we need to keep up with the latest tooling in the open source world.

This is not directly related to upgrading to Babel 7, but our ESLint parser is `babel-eslint`, so fresh dependencies can get us to the latest versions of the packages that make up our ESLint configuration, which is exposed by `eslint-config-kyt`. We can also add specific rules to this config via `eslint-plugin-kyt`, which has been added during this upgrade.

## Building the App
When building our application, we need to make sure we are taking advantage of all optimizations recommended for running React in production. `babel-preset-kyt-react` now includes all React-specific Babel transforms that we think are suitable for all apps running React. For optimizations on prod, we include the following plugins, in addition to `@babel/preset-react`:

```
@babel/plugin-transform-react-constant-elements
@babel/plugin-transform-react-inline-elements
babel-plugin-add-react-displayname
babel-plugin-transform-react-remove-prop-types
```

`react-constant-elements` and `react-inline-elements` can be turned off by passing `useProductionTransforms: false` to `babel-preset-kyt-react`.

The KYT Core preset provides the `includeRuntime` flag, which further allows us to optimize our production bundles. By turning this flag on in our `.babelrc`, we can drastically reduce the size of our production JS bundles by replacing Babel transform output with references to Babel helper functions. These functions will add to the size our vendor bundle modestly, while reducing the size of our individual route bundles dramatically - eg. the "story" bundle size on NYT is reduced by over 200KB. All consuming apps need to specify `@babel/polyfill` and `@babel/runtime` as dependencies.

It is possible that a new package is created called "kyt-runtime" which will simply include all of these explicit requirements.

### Running the app in development
Nothing should change with how the app runs while developing, as KYT has not experienced any substantial changes as pertains to Webpack - updates to Webpack will happen later.

### Running the app in production
The consuming app should be able to run without errors after `npm prune --production` is invoked. This means that any runtime dependencies should exist in the "dependencies" portion of `package.json`. Anything included in `kyt` will be pruned, as it is a dev dependency, hence the need for a convenience package like `kyt-runtime` to take care of any production deps, regardless of their inclusion in `kyt`.


packages/kyt-core/package.json
```diff
@@ -62,23 +63,24 @@
...
+    "webpack-dev-middleware": "2.0.6",
+    "webpack-dev-server": "2.11.3",
+    "webpack-hot-middleware": "2.24.3",
+    "webpack-merge": "4.1.5",
+    "webpack-node-externals": "1.7.2",
+    "webpack-plugin-hash-output": "1.3.1"
   },
   "peerDependencies": {
-    "babel-polyfill": "6.26.0"
+    "kyt-runtime": "pre"
```
delambo: Should we make this "beta" ahead of the merge to master and then publish a beta? Or do you want to do it in a follow up?

PR #586, opened 2019-01-31
Upgrade to Jest 24, ditch babel-core

staylor:


staylor:
@fsouza this is a known failure that we are ok with throughout the Alpha dev cycle here. will address before we bump real versions

PR #591, opened 2019-02-23
Separate runtime chunks by entry, update starter kyts to use latest

staylor:
`optimization.runtimeChunk` should create a runtime chunk per entry. This requires some updates to the starter kyts, so I updated them to also use `dynamic()` and React Router v4


staylor:
packages are built as: `1.0.0-alpha.43822f6a`

e2e_tests/tests/kyt-build.test.js
```diff
@@ -20,7 +20,7 @@ describe('kyt build', () => {
     expect(shell.test('-f', 'build/public/nothing.txt')).toBe(true);
 
     // Should produce the manifest, main and vendor scripts
-    expect(shell.ls('build/public/manifest-*.js').code).toBe(0);
+    expect(shell.ls('build/public/runtime~main-*.js').code).toBe(0);
```
tizmagik: We used to inline this internally (`KYT.ASSETS_MANIFEST`) -- would that still be necessary/possible?

staylor: Yes - can still be plucked out by clientAssets[‘runtime~main.js’], I’ll make a Vi patch to illustrate

PR #595, opened 2019-03-29
Add Terser webpack plugin and remove UglifyJS for minification

ccincotti3:
UglifyJS React hooks issue reported here - https://github.com/facebook/react/issues/14014

Switching to terser-webpack-plugin


delambo:
Going to close this in favor of just removing the minimizer #600. Thanks for leading on this change @ccincotti3

PR #608, opened 2019-10-01
bring react-loadable in-house, remove old React APIs

staylor:
`react-loadable` hasn't really been updated in over a year, seems abandoned. That's fine, the API is not too complicated.

In this PR:
* use `React.createContext()` instead of the old `getChildContext` technique
* stop using `componentWillMount()` lifecycle method
* consume context in `LoadableComponent` via `static contextType`
* delete `LoadableMap` code, we don't want it
* Add unit tests for `kyt-starter-universal` to test Loadable components

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -81,15 +81,14 @@
       }
     },
     "@babel/generator": {
-      "version": "7.5.5",
-      "resolved": "https://registry.npmjs.org/@babel/generator/-/generator-7.5.5.tgz",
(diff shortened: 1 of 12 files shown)


staylor:
the e2e tests have a warning because `webpack-node-externals` does not look backwards when excluding `node_modules` - the tests run the starter kyt folder, which has some symlinked deps and resolves its other deps by walking to the root

PR #615, opened 2019-12-09
This file is not compiled, can break IE11

staylor:


packages/kyt-runtime/client.js
```diff
@@ -1,3 +1,3 @@
-const { preloadReady } = require('./lib/loadable');
+var { preloadReady } = require('./lib/loadable');
```
Juan-M: I don't think the destructuring is going to get digested either.
I'll open another PR to fix it.

Juan-M: Yeah IE is throwing another error: expected identifier.

PR #616, opened 2019-12-10
STRIK-1008: fully remove ES6 from client.js

Juan-M:
https://github.com/nytimes/kyt/pull/615 left behind the destructuring of the module that makes IE keep tripping.


Juan-M:
@staylor It seems I can't assign reviewers. I don't seem to have push access to the repo so I made the PR from a fork. I'm not sure if it's going to cause more issues. It's been a while since I did this for the last time.

PR #643, opened 2020-03-17
Update `kyt dev` CLI action

staylor:
Moves our `dev` command closer to projects like Razzle and Create React App. Removes all use of React Hot Loader. A lot of the tasks our `dev` command performs are similar to what already exist in `react-dev-utils`.


packages/kyt-core/src/actions/__tests__/dev.test.js
```diff
@@ -37,45 +24,19 @@ const dev = require('../dev');
...
-      ifPortIsFreeDo,
-      webpackCompiler,
-      buildConfigs,
-    ].forEach(mock => mock.mockClear());
+    jest.clearAllMocks();
   });
 
   const href = { href: 'href' };
-  const port = 1000;
+  const port = 1024;
```
woodb: I know it doesn't matter in this test, but just curious why the port change here? Double check that the test fails or something?

packages/kyt-core/src/config/webpack.dev.server.js
```diff
@@ -2,25 +2,51 @@
...
+  clientAssetsFile,
+  loadableAssetsFile,
+} = require('kyt-utils/paths')();
 const getPolyfill = require('./getPolyfill');
 
+const nodeArgs = ['-r', 'source-map-support/register'];
+// Passthrough --inspect and --inspect-brk flags (with optional [host:port] value) to node
+if (process.env.INSPECT_BRK) {
+  nodeArgs.push(process.env.INSPECT_BRK);
+} else if (process.env.INSPECT) {
```
woodb: Should we warn if someone supplies both flags?

packages/kyt-core/src/config/webpack.dev.server.js
```diff
@@ -32,10 +58,16 @@ module.exports = options => ({
...
-      raw: true,
-      entryOnly: true,
+    // Prevent creating multiple chunks for the server
+    new webpack.optimize.LimitChunkCountPlugin({
+      maxChunks: 1,
+    }),
+    new webpack.HotModuleReplacementPlugin(),
+    new StartServerPlugin({
+      name: 'main.js',
+      nodeArgs,
```
woodb: Where was this before? Are we sure we wanna introduce a `main.js` requirement?

packages/kyt-core/src/config/webpackHotDevClient.js
```diff
@@ -0,0 +1,273 @@
+// This alternative WebpackDevServer combines the functionality of:
+// https://github.com/webpack/webpack-dev-server/blob/webpack-1/client/index.js
+// https://github.com/webpack/webpack/blob/webpack-1/hot/dev-server.js
+
+// It only supports their simplest configuration (hot updates on same server).
+// It makes some opinionated choices on top, like adding a syntax error overlay
+// that looks similar to our console output. The error overlay is inspired by:
+// https://github.com/glenjamin/webpack-hot-middleware
+
+/* eslint-disable no-use-before-define */
```
woodb: I presume this can't be avoided here

packages/kyt-core/src/config/webpackHotDevClient.js
```diff
@@ -0,0 +1,273 @@
...
+
+const SockJS = require('sockjs-client');
+const stripAnsi = require('strip-ansi');
+const url = require('url');
+const launchEditorEndpoint = require('react-dev-utils/launchEditorEndpoint');
+const formatWebpackMessages = require('react-dev-utils/formatWebpackMessages');
+const ErrorOverlay = require('react-error-overlay');
+
+const sockJSPort = KYT.CLIENT_PORT || parseInt(window.location.port || 3000, 10) + 1;
```
woodb: Not sure how important this is, but `KYT.CLIENT_PORT` will be a String, so this will be a String or Number depending on whether that's defined

packages/kyt-core/src/config/webpackHotDevClient.js
```diff
@@ -0,0 +1,273 @@
...
+  }
+}
+
+// Compilation with warnings (e.g. ESLint).
+function handleWarnings(warnings) {
+  clearOutdatedErrors();
+
+  const isHotUpdate = !isFirstCompilation;
+  isFirstCompilation = false;
+  hasCompileErrors = false;
```
woodb: Why do we set this to `false` here? Will we simply not get to this line of code if there are compile errors?

packages/kyt-core/src/config/webpackHotDevClient.js
```diff
@@ -0,0 +1,273 @@
...
+  })
+);
+
+// Unlike WebpackDevServer client, we won't try to reconnect
+// to avoid spamming the console. Disconnect usually happens
+// when developer stops the server.
+connection.onclose = () => {
+  if (typeof console !== 'undefined' && typeof console.info === 'function') {
+    console.info('The development server has disconnected.\nRefresh the page if necessary.');
+  }
```
woodb: We may wanna make this louder than an `info`
```suggestion
  if (typeof console !== 'undefined' && typeof console.warn === 'function') {
    console.warn('⚠️ The development server has disconnected.\nRefresh the page if necessary.');
  }
```

packages/kyt-core/src/config/webpackHotDevClient.js
```diff
@@ -0,0 +1,273 @@
...
+  // https://webpack.github.io/docs/hot-module-replacement.html#check
+  const result = module.hot.check(/* autoApply */ true, handleApplyUpdates);
+
+  // // Webpack 2 returns a Promise instead of invoking a callback
+  if (result && result.then) {
+    result.then(
+      updatedModules => {
+        handleApplyUpdates(null, updatedModules);
+      },
+      err => {
```
woodb: Should this be in a `.catch(...`?

packages/kyt-core/src/config/webpackHotDevClient.js
```diff
@@ -0,0 +1,273 @@
...
+      // While we were updating, there was a new update! Do it again.
+      tryApplyUpdates();
+    }
+  }
+
+  // https://webpack.github.io/docs/hot-module-replacement.html#check
+  const result = module.hot.check(/* autoApply */ true, handleApplyUpdates);
+
+  // // Webpack 2 returns a Promise instead of invoking a callback
+  if (result && result.then) {
```
woodb:
```suggestion
  if (result instanceof Promise) {
```

packages/kyt-core/src/config/webpackHotDevClient.js
```diff
@@ -0,0 +1,273 @@
...
+    result.then(
+      updatedModules => {
+        handleApplyUpdates(null, updatedModules);
+      },
+      err => {
+        handleApplyUpdates(err, null);
+      }
+    );
+  }
```
woodb: It doesn't seem like we handle the callback version here

packages/kyt-core/src/config/webpack.dev.server.js
staylor: previously, this was an extra Express server in code - all of that logic has been removed, simplifies the `dev` action considerably. This is what Razzle does: https://github.com/jaredpalmer/razzle/blob/master/packages/razzle/config/createConfig.js#L337

packages/kyt-core/src/config/webpackHotDevClient.js
staylor: it cannot, this is mostly a copy of: https://github.com/jaredpalmer/razzle/blob/master/packages/razzle-dev-utils/webpackHotDevClient.js

packages/kyt-core/src/config/webpackHotDevClient.js
staylor: just pushed a parseInt() for this

packages/kyt-core/src/config/webpackHotDevClient.js
staylor: the flag gets reset on hot compilations

packages/kyt-core/src/config/webpackHotDevClient.js
staylor: I just dropped this block, which was only there to support Webpack 2

packages/kyt-core/src/config/webpackHotDevClient.js
staylor: I just dropped this block, which was only there to support Webpack 2

packages/kyt-core/src/actions/__tests__/dev.test.js
staylor: the port check is not mocked, and I think 1000 is invalid

packages/kyt-core/src/config/webpack.dev.server.js
staylor: the dev entrypoint *is* `main.js`

packages/kyt-core/src/config/webpackHotDevClient.js
staylor: I just dropped this block, which was only there to support Webpack 2

packages/kyt-core/src/config/webpack.dev.server.js
staylor: copying this: https://github.com/jaredpalmer/razzle/blob/ef3041789b2c9c56c602991e9383e59e483cc666/packages/razzle/config/createConfig.js#L325

let's not worry about it for now

staylor:
build passes in Travis CI, just not reported yet: https://travis-ci.org/github/nytimes/kyt/jobs/666939985

PR #648, opened 2020-04-13
add memory flag, and add hot accept to starter

staylor:


packages/kyt-core/src/config/webpack.dev.server.js
```diff
@@ -11,7 +11,7 @@ const {
 const StartServerPlugin = require('./StartServerPlugin');
 const getPolyfill = require('./getPolyfill');
 
-const nodeArgs = ['-r', 'source-map-support/register'];
+const nodeArgs = ['-r', 'source-map-support/register', '--max_old_space_size=4096'];
```
woodb, line 14: Not sure if it really likes it this way or not, I'm guessing not but jic, here it is
```suggestion
const nodeArgs = ['-r', 'source-map-support/register', '--max_old_space_size', '4096'];
```

packages/kyt-core/src/config/webpack.dev.client.js
```diff
@@ -63,11 +63,6 @@ module.exports = options => {
       },
     },
 
-    plugins: [
-      ...kytWebpackPlugins(options),
-      new webpack.HotModuleReplacementPlugin({
-        multiStep: true,
-      }),
-    ],
+    plugins: [...kytWebpackPlugins(options), new webpack.HotModuleReplacementPlugin()],
```
woodb, line 66: Will the removal of the `multistep` make reloads slower?

PR #785, opened 2021-05-20
fix(runtme): check for null bundle

staylor:


packages/kyt-runtime/src/server.js
```diff
@@ -47,7 +47,9 @@ export const getBundles = ({ entry = 'main', modules, assets = null, loadableBun
 
   const jsBundles = bundles.filter(
     b =>
-      b.file.endsWith('.js') && !b.file.endsWith('.hot-update.js') && !hashes.includes(b.publicPath)
+      b?.file?.endsWith('.js') &&
+      !b.file.endsWith('.hot-update.js') &&
```
fesebuv, line 51: if `b` is `undefined` wouldn't this throw a `TypeError` as well?
```!b.file.endsWith('.hot-update.js') ```

fesebuv: nvm this should be fine.

PR #787, opened 2021-05-20
fix: concat not push

staylor:


jonahsmith (approved):

john-michael-murphy (approved):

BenjaminCheung (approved):

packages/kyt-runtime/src/loadable-plugin.js
```diff
@@ -119,9 +119,6 @@ export class LoadablePlugin {
 
 export function getBundles(manifest, moduleIds) {
   return moduleIds.reduce((bundles, moduleId) => {
-    if (manifest[moduleId]) {
-      bundles.push(manifest[moduleId]);
-    }
-    return bundles;
+    return manifest[moduleId] ? bundles.concat(manifest[moduleId]) : bundles;
```
fesebuv, line 122: We should include unit tests here.

PR #1075, opened 2022-09-28
feat(babel-preset-kyt-react): add babel-plugin-pure-static-props when productionTransforms is true

kohlmannj:
relates to [DSYS-10](https://jira.nyt.net/browse/DSYS-10)

## Description

This PR adds [`babel-plugin-pure-static-props`](https://github.com/mbrowne/babel-plugin-pure-static-props) to `babel-preset-kyt-react`'s set of production transforms (enabled when the preset's `productionTransforms` option is set to `true`).

## Motivation / Background

Static properties on [React] components break tree-shaking: https://github.com/styled-components/babel-plugin-styled-components/issues/245#issuecomment-525844957

This same commenter created `babel-plugin-pure-static-props` to address this. From its [README](https://github.com/mbrowne/babel-plugin-pure-static-props):  

> This plugin replaces static property assignments on React components (e.g. `MyComponent.defaultProps = {...}`) with `Object.assign()` statements annotated with `/*#__PURE__*/` comments so that tree-shaking will work correctly.

This is an especially important optimization for `babel-preset-kyt-react` because we add the static property `displayName` to every React component via [`babel-plugin-add-react-displayname`](https://babeljs.io/docs/en/babel-plugin-transform-react-display-name). That is, this plugin is always enabled. (For that reason, we may consider always enabling `babel-plugin-pure-static-props` as well.)

### DIFF
--- a/packages/babel-preset-kyt-react/package.json
+++ b/packages/babel-preset-kyt-react/package.json
@@ -17,6 +17,7 @@
     "@babel/plugin-transform-react-inline-elements": "7.16.7",
     "@babel/preset-react": "7.14.5",
     "babel-plugin-add-react-displayname": "0.0.5",
+    "babel-plugin-pure-static-props": "0.2.0",
     "babel-plugin-transform-react-remove-prop-types": "0.4.24",
     "babel-preset-kyt-core": "2.0.0"
   },
--- a/packages/babel-preset-kyt-react/src/index.js
+++ b/packages/babel-preset-kyt-react/src/index.js
@@ -19,11 +19,10 @@ module.exports = function getPresetReact(context, opts) {
   if (useProductionTransforms === true) {
     productionTransforms.push(reactTransformConstant);
     productionTransforms.push(reactTransformInline);
-    productionTransforms.push(pureStaticProps);
   }
 
   return {
-    plugins: [addReactDisplayName],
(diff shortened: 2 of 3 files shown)


rmfarrell (approved):

delambo (approved):

kohlmannj:
I took a moment to test this in the Monorepo. I found that nytimes/new#517 unconditionally sets `useProductionTransforms: false` for all Monorepo libraries that use `babel-preset-kyt-react`.

Thus, I am going to revise this PR to unconditionally enable `babel-plugin-pure-static-props`, as previously hinted at in the original PR description.

Meanwhile, I feel confident that, in addition to more local testing, a corresponding Monorepo PR after merging and releasing _this_ PR will be enough to tell us whether it's safe to unconditionally enable this plugin.

TL;DR — One change coming!

kohlmannj:
Turns out I might be closing this PR. [These test failures](https://github.com/nytimes/kyt/actions/runs/3146733930/jobs/5115511969) are representative of an issue I encountered with at least one source file in the Monorepo. I have no idea what to do about it, as the issue seems to lie within the plugin itself. And diagnosing that issue, while potentially valuable in the long term, moves this from a "quick fix" (just add plugin) to "potentially more involved fix."

Anyway, closing for now since unconditionally enabling `babel-plugin-pure-static-props` could cause build errors.

PR #1173, opened 2024-02-28
fix: support custom url protocol scheme on websockets

ilyaGurevich:
Custom url protocol schemes are not supported in sockjs, this allows you to render kyt-generated content locally in custom URL patterns by using a web-socket connection instead.

### DIFF
--- a/packages/kyt-core/src/config/webpackHotDevClient.js
+++ b/packages/kyt-core/src/config/webpackHotDevClient.js
@@ -57,16 +57,31 @@ if (module.hot && typeof module.hot.dispose === 'function') {
   });
 }
 
-// Connect to WebpackDevServer via a socket.
-const connection = new SockJS(
-  url.format({
-    protocol: window.location.protocol,
-    hostname: window.location.hostname,
-    port: sockJSPort,
-    // Hardcoded in WebpackDevServer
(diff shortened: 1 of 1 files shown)


packages/kyt-core/src/config/webpackHotDevClient.js
```diff
@@ -57,16 +57,31 @@ if (module.hot && typeof module.hot.dispose === 'function') {
...
-  url.format({
-    protocol: window.location.protocol,
-    hostname: window.location.hostname,
-    port: sockJSPort,
-    // Hardcoded in WebpackDevServer
-    pathname: '/sockjs-node',
-  })
-);
+let connection;
+if (!window.location.protocol.includes('http')) {
```
kmnaid, line 61: Can this protocol be `https`?

ilyaGurevich: If you're `https` then the `.includes` will also match with `https`

ilyaGurevich: const protocol = 'https:'
protocol.includes('http') --> true

kmnaid (approved):

delambo (approved):

PR #1174, opened 2024-04-08
add css/scss modules support, back

delambo:
- adds support for mise and node 20
- adds support for CSS and SCSS Modules, file names: `*.module.scss`
- adds long overdue front page deprecation notice


.github/workflows/continuous-integration.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [12.x, 14.x, 16.x]
+        node-version: [18.x, 20.x]
```
lquixada, line 17: 18.x?

.mise.toml
```diff
@@ -0,0 +1,9 @@
+[tools]
+node = '20.10.0'
+# stuck on this version of yarn because of some weird problems
+# related to babel tooling when installing new dependencies
+yarn = '1.19.0'
+
+[env]
+RTX_FETCH_REMOTE_VERSIONS_TIMEOUT="300 s"
```
lquixada, line 8: is there a mise alternative for this variable?

lquixada (approved):
added couple of comments. nothing blocking.

.github/workflows/continuous-integration.yml
delambo: I added support for 20. That includes support for 18. So why not 🤷?

.mise.toml
delambo: I think so, but it's not documented very well. I updated -- thanks for pointing that out.

(This timeout is needed because node.org can have excessively slow download times. Not sure if that's fixed at this point.)

